PR #1, opened 2024-06-12
Add `package:record_use`

mosuem:
Add `package:record_use`. This is part of an effort to allow tree-shaking of non-Dart objects, such as data files or native code.

---

<details>

</details>


.github/workflows/record_use.yaml
```diff
@@ -0,0 +1,39 @@
+name: record_use
```
devoncarew, line 1:
```suggestion
name: 'package:record_use'
```

pkgs/record_use/README.md
```diff
@@ -0,0 +1,115 @@
...
+being recorded.
+
+- If placed on a static method, the annotation means that arguments passed to
+the method will be recorded, as far as they can be inferred at compile time.
+- If placed on a class with a constant constructor, the annotation means that
+any constant instance of the class will be recorded. This is particularly useful
+when using the class as an annotation.
+
+## Example
+```dart
```
devoncarew, line 22:
```suggestion

```dart
```

pkgs/record_use/README.md
```diff
@@ -0,0 +1,115 @@
...
+    ... // Do something with the information, such as tree-shaking native assets
+  });
+}
+```
+
+## Installation
+To install the record_use package, run the following command:
+
+```bash
+dart pub add record_use
```
devoncarew, line 98: Typically (always?) this package will be added as a regular dependency?

pkgs/record_use/lib/src/proto/usages_read.proto
```diff
@@ -0,0 +1,44 @@
+syntax = "proto3";
```
devoncarew, line 1: If this is hand-written it'll need a copyright header.

pkgs/record_use/lib/src/data_classes/extensions.dart
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:typed_data';
+
+import '../proto/usages_read.pb.dart' as pb;
```
devoncarew, line 8: Having all consumers of this package transitively take a dep on `package:protobuf` is a huge consideration. protobuf here is used to facilitate writing and reading usage information to disk? It that file format intended to be canonical (can the annotation be used meaningfully without understanding this file format)?

pkgs/record_use/lib/src/proto/usages_storage.proto
```diff
@@ -0,0 +1,45 @@
+syntax = "proto3";
```
devoncarew, line 1: copyright header

devoncarew (commented):
My main comment here is around the use of protos. Introducing a transitive dep on that package to consumers is a large consideration. Additionally, the format is not typically used outside of Google specific projects (grpc, ...).

Are we expecting the files to be very large? The format to be difficult to encode / decode? Is evolving the format in backwards compatible ways important?

How would the tradeoffs of encoding to and from json compare?

mosuem:
> My main comment here is around the use of protos. Introducing a transitive dep on that package to consumers is a large consideration. Additionally, the format is not typically used outside of Google specific projects (grpc, ...).

This package will be a de-facto dev dependency, as it is only being used in link hooks. So the size should not be an issue, right?

> Are we expecting the files to be very large? The format to be difficult to encode / decode? Is evolving the format in backwards compatible ways important?

Evolving the format and size are the two main drivers to go for protobuf, together with the ability to easily understand the format. The size could be large, depending on how user code looks like in practice, but the file would be only used at build time, so has no implications on released app size.

> How would the tradeoffs of encoding to and from json compare?

JSON has a more difficult migration path for adding/removing support for keys, and is slightly larger in storage. It is easier to manually inspect stored files, which is why we would use the `toDebugString()` method to output debug files for `verbose` compilation modes. Protobuf has a schema which is simpler than JSON, and has better support for types.

I am fine with using JSON, but do like the easier migration with protobuf for future changes.

pkgs/record_use/README.md
mosuem: This was copied from a generic package readme - I think the only package using this will be `native_assets_cli`...

pkgs/record_use/lib/src/proto/usages_read.pbenum.dart
```diff
@@ -0,0 +1,10 @@
+//
+//  Generated code. Do not modify.
+//  source: usages_read.proto
+//
+// @dart = 2.12
```
devoncarew, line 5: Note to self: `2.12` is now considered an old version of Dart; we should update the proto generator to use at leat `3.0`.

pkgs/record_use/README.md
```diff
@@ -0,0 +1,116 @@
+# `package:record_use`
+
+
+> [!CAUTION]
+> This is an experimental package, and it's API can break at any time. Use at
+> your own discretion.
+
+This package provides the data classes for the usage recording feature in the
+Dart SDK.
+
```
devoncarew, line 10: You might want to say something like 'this package is the canonical way to read usage files written by the sdk' (`RecordedUsages.fromFile`); if this is the main way to read those files, that could give us a bit more flexibility in terms of evolving the format.

pkgs/record_use/example/record_use_example.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:record_use/record_use.dart';
+
+void doStuff(RecordedUsages usage, Identifier callId, Identifier referenceId) {
```
devoncarew, line 7: Do you want to provide an example of using `RecordedUsages.fromFile`? And / or borrow the main example from the readme.

devoncarew (approved):
lgtm and no more comments re: the file format. If we do find that this package is overly constraining the ecosystem re: the protobuf dep. / leaking the dep transitively to more places than expected we should be open to reexamining this choice.

mkustermann (commented):
I think this will need some iteration, but tto get it started: lgtm with comments

Though I think it makes more sense for this package to live together with `dart-lang/native`: The users of this package will be `hook/link.dart` scripts which already depend on other packages in `dart-lang/native`.

@mosuem  Could we land it there?

pkgs/record_use/README.md
```diff
@@ -0,0 +1,116 @@
...
+  const RecordMetadata(this.metadata);
+}
+
+```
+This code will generate a data file that contains both the `metadata` values of
+the `RecordMetadata` instances, as well as the arguments for the different
+methods annotated with `@RecordUse()`.
+
+This information can then be accessed in a link hook as follows:
+```dart
```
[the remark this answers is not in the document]
mosuem: I believe they should be in the example and in the readme, it's nicer for users to see it at a glance.

pkgs/record_use/lib/src/proto/usages_storage.proto
```diff
@@ -0,0 +1,49 @@
...
+message Location {
+  uint32 uri = 1; 
+  uint32 line = 2;
+  uint32 column = 3;
+}
+
+message Reference {
+  Location location = 1;
+  string loading_unit = 2;
+  oneof reference {
```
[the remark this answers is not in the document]
mosuem: I believe that the usages have much in common:
- A definition the annotation is placed on (a class or a method)
- Additional information which is also recorded (arguments or fields)

The annotation also denotes a central concept: Record usages of the object in the code.

While there are also some differences, I do like this common "interface".

pkgs/record_use/example/record_use_example.dart
mosuem: I think `RecordedUsages.fromFile` is an internal method/implementation detail, the users will get a deserialized `RecordUsages` object in their `link.dart`.

pkgs/record_use/README.md
mosuem: This is the only way to read the files!

Issue #366, opened 2023-12-05
Async Cache is caching exceptions


akmalviya03:
### Steps to reproduce

To reproduce the same issue you can refer the code sample given below.

### Expected results

It should only cache successful result.

### Actual results

It should not cache exception.

```
This callback is running
We have successfully cancelled future
Error First Exception: Future Cancelled
Error Second Exception: Future Cancelled
```

### Code sample

<details><summary>Code sample</summary>

```dart
import 'package:async/async.dart';

late AsyncCache<String> temporaryCache;
late CancelableOperation<String> cancelableOperation;

Future<void> main(List<String> arguments) async {
  temporaryCache = AsyncCache(Duration(seconds: 15));
  cancelableOperation =
      CancelableOperation.fromFuture(_futureToRun(), onCancel: () {
    print('We have successfully cancelled future');
  });
   getData().then((value) => print('First')).onError((error, stackTrace) => print('Error First $error'));
  cancelableOperation.cancel();
  getData().then((value) => print('Second')).onError((error, stackTrace) => print('Error Second $error'));
}

Future<String> _futureToRun() async {
  await Future.delayed(const Duration(seconds: 5));
  return Future.value("Temporary Cache");
}

Future<String> getData() async {
  return temporaryCache.fetch(() async {
    print('This callback is running');
    String? value = await cancelableOperation.valueOrCancellation();
    if (value != null) {
      return value;
    }
    throw Exception('Future Cancelled');
  });
}

```
</details>

PR #548, opened 2024-10-17
fix: async cache storing exception fixed

akmalviya03:
Earlier async cache was caching exceptions.
With new changes in `fetch` method of `AsyncCache` class will no longer store exception.
By default `AsyncCache` will store exception, by setting `_canCacheException` to false. It won't store exception.

Please refer dart-lang/core#366 for video demo of both issue and solution along with minimal reproducible code.

---


pkgs/async/test/async_cache_test.dart
```diff
@@ -18,6 +18,20 @@ void main() {
...
+    cache = AsyncCache(const Duration(hours: 1), cacheErrors: false);
+
+    Future<String> asyncFunctionThatThrows() {
+      throw Exception();
+    }
+
+    var errorThrowingFuture = cache.fetch(asyncFunctionThatThrows);
+    await expectLater(errorThrowingFuture, throwsA(isException));
+
+    var valueFuture = cache.fetch(() async => 'Success');
```
lrhn, line 31: Consider checking that time hasn't progressed by one hour. Just for good measure.

pkgs/async/test/async_cache_test.dart
```diff
@@ -18,6 +18,20 @@ void main() {
     cache = AsyncCache(const Duration(hours: 1));
   });
 
+  test('should not fetch when callback throws exception', () async {
```
lrhn, line 21: Not sure "fetch" is the verb I'd use here (aka: I don't know what the description means).
It's consistent with the other tests. I also can't read those.

That's not really *your* issue to fix, though. (Should we say "recompute" instead of "fetch". Why is the method called `fetch` to begin with, seems to assume a specific use-case.)

lrhn (approved):

pkgs/async/test/async_cache_test.dart
akmalviya03: Initially I also thought the same to rename it to something more generic.

pkgs/async/test/async_cache_test.dart
akmalviya03: Done updated test cases.

mosuem:
@lrhn i think it's your turn for the review ;) Just a friendly ping!

PR #577, opened 2024-10-17
Update tables to Unicode 16.0.0.

lrhn:
Updates tables to be generated from Unicode 16.0.0.

Addd license file change check. (Since the license file changed this time,
and it's something that one should check if it happens.)

Some clean-up and typo-fixes.

Documents that the BG9c rule from Unicode 15.10 is not supported. 
(That would require more bits of information per character, information that the current table
builder does not have, likely doubling this size of the character property table when each character's
category no longer fits in 4 bits, and will require a correspondingly larger state machine with more states times more input character categories.)


pkgs/characters/pubspec.yaml
```diff
@@ -12,5 +12,5 @@ environment:
   sdk: ^3.4.0
 
 dev_dependencies:
-  dart_flutter_team_lints: ^3.0.0
+  dart_flutter_team_lints: ^3.2.0
```
devoncarew, line 15: From the CI error - `Because characters depends on dart_flutter_team_lints >=3.2.0 which requires SDK version >=3.5.0 <4.0.0, version solving failed` - I'd leave this dep at ^3.0.0:

```suggestion
  dart_flutter_team_lints: ^3.0.0
```

lrhn: Bigger isn't better, who'd have thunk? (At least we know we're prepared.)

pkgs/characters/lib/src/characters.dart
```diff
@@ -19,7 +19,7 @@ import "characters_impl.dart";
 /// which allows iterating the independent characters in both directions,
 /// but which also provides ways to select other ranges of characters
 /// in different ways.
-abstract class Characters implements Iterable<String> {
+abstract interface class Characters implements Iterable<String> {
```
natebosch, line 22: Can we move this to a separate commit? Or if not, please include a note about this change in the commit message. I think we should also include it in the changelog.

natebosch (approved):

pkgs/characters/lib/src/characters.dart
lrhn: Will revert for now.
It's not like it's *useful* to extend an empty class, but people do ... things.

PR #713, opened 2024-10-25
os_detect: bump lints and SDK versions

kevmoo:
...and other cleanup across the board

- drop deprecated lints
- drop redundant lints
- tiny bump to min test version using --tighten pub command

### DIFF
--- a/pkgs/args/pubspec.yaml
+++ b/pkgs/args/pubspec.yaml
@@ -13,4 +13,4 @@ environment:
 
 dev_dependencies:
   dart_flutter_team_lints: ^3.0.0
-  test: ^1.16.0
+  test: ^1.16.6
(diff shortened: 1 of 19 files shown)


pkgs/args/pubspec.yaml
```diff
@@ -13,4 +13,4 @@ environment:
 
 dev_dependencies:
   dart_flutter_team_lints: ^3.0.0
-  test: ^1.16.0
+  test: ^1.16.6
```
devoncarew, line 16: Just curious why the increase in specificity?

devoncarew (approved):

pkgs/args/pubspec.yaml
kevmoo: It's what `dart pub upgrade --tighten` does 🤷

PR #715, opened 2024-11-01
Introduce `IterableMapEntryExtension` for use with `Map.entries`.

jonasfj:
These utilities could often be useful when working with maps. We could also put this into a different utility package.

**Example**:
```dart
final myMap = {
  'foo': 42,
  'bar': -1,
  'foobar': 21,
};

// myMap without negative values
myMap.entries.whereValue((v) => v >= 0).toMap();

// myMap, but only keys that start with 'foo'
myMap.entries.whereKey((k) => k.startsWith('foo')).toMap();
```

------

I'm tempted to also suggest:
```dart
  Iterable<T> mapKeyValue<T>(T Function(K, V) convert) =>
      map((e) => convert(e.key, e.value));
```

But the name is a bit long. `mapKV` is lovely short, but probably too ugly.
When doing `.map` over `Iterable<MapEntry>` you generally don't want to write `.map((entry) => entry.key`.

### DIFF
--- a/pkgs/collection/lib/src/iterable_extensions.dart
+++ b/pkgs/collection/lib/src/iterable_extensions.dart
@@ -914,6 +914,33 @@ extension IterableIterableExtension<T> on Iterable<Iterable<T>> {
       };
 }
 
+/// Extension on iterables of [MapEntry].
+///
+/// An [Iterable<MapEntry>] is obtained using [Map.entries], these extensions
+/// make it easy to work on a [Map] as a list of pairs.
+extension IterableMapEntryExtension<K, V> on Iterable<MapEntry<K, V>> {
+  /// Creates a new lazy [Iterable] with all elements whose [MapEntry.key]
+  /// satisfy the predicate [test].
+  Iterable<MapEntry<K, V>> whereKey(bool Function(K) test) =>
+      where((e) => test(e.key));
+
+  /// Creates a new lazy [Iterable] with all elements whose [MapEntry.value]
+  /// satisfy the predicate [test].
+  Iterable<MapEntry<K, V>> whereValue(bool Function(V) test) =>
+      where((e) => test(e.value));
+
+  /// Create an new lazy [Iterable] with [MapEntry.key] from all elements.
+  Iterable<K> get keys => map((e) => e.key);
+
+  /// Create an new lazy [Iterable] with [MapEntry.value] from all elements.
+  Iterable<V> get values => map((e) => e.value);
+
+  /// Create a [Map] from all elements.
+  ///
+  /// This is a short-hand for [Map.fromEntries].
+  Map<K, V> toMap() => Map.fromEntries(this);
+}
+
 /// Extensions that apply to iterables of [Comparable] elements.
 ///
 /// These operations can assume that the elements have a natural ordering,
--- a/pkgs/collection/lib/src/iterable_extensions.dart
+++ b/pkgs/collection/lib/src/iterable_extensions.dart
@@ -916,29 +916,39 @@ extension IterableIterableExtension<T> on Iterable<Iterable<T>> {
 
 /// Extension on iterables of [MapEntry].
 ///
-/// An [Iterable<MapEntry>] is obtained using [Map.entries], these extensions
-/// make it easy to work on a [Map] as a list of pairs.
+/// An [Iterable<MapEntry>] is obtained using [Map.entries]. These extensions
+/// facilitates working directly on the entries of a [Map].
 extension IterableMapEntryExtension<K, V> on Iterable<MapEntry<K, V>> {
-  /// Creates a new lazy [Iterable] with all elements whose [MapEntry.key]
-  /// satisfy the predicate [test].
+  /// The elements whose [MapEntry.key] values satisfy [test].
+  ///
+  /// The resulting iterable is lazily computing its elements
+  /// based on the elements this iterable.
   Iterable<MapEntry<K, V>> whereKey(bool Function(K) test) =>
       where((e) => test(e.key));
 
-  /// Creates a new lazy [Iterable] with all elements whose [MapEntry.value]
-  /// satisfy the predicate [test].
+  /// The elements whose [MapEntry.value] values satisfy [test].
+  ///
+  /// The resulting iterable is lazily computing its elements
+  /// based on the elements this iterable.
   Iterable<MapEntry<K, V>> whereValue(bool Function(V) test) =>
       where((e) => test(e.value));
 
-  /// Create an new lazy [Iterable] with [MapEntry.key] from all elements.
+  /// A new lazy [Iterable] of the [MapEntry.key]s of these entries.
+  ///
+  /// Do not use this getter as `map.entries.keys`, just use `map.keys`
+  /// directly.
   Iterable<K> get keys => map((e) => e.key);
 
-  /// Create an new lazy [Iterable] with [MapEntry.value] from all elements.
+  /// A new lazy [Iterable] of the [MapEntry.value]s of these entries.
+  ///
+  /// Do not use this getter as `map.entries.values`, just use `map.values`
+  /// directly.
   Iterable<V> get values => map((e) => e.value);
 
-  /// Create a [Map] from all elements.
+  /// Create a [Map<K, V>] from all elements.
   ///
   /// This is a short-hand for [Map.fromEntries].
-  Map<K, V> toMap() => Map.fromEntries(this);
+  Map<K, V> toMap() => Map<K, V>.fromEntries(this);
 }
 
 /// Extensions that apply to iterables of [Comparable] elements.
--- a/pkgs/collection/pubspec.yaml
+++ b/pkgs/collection/pubspec.yaml
@@ -1,5 +1,5 @@
 name: collection
-version: 1.19.1
+version: 1.19.1-wip
 description: >-
   Collections and utilities functions and classes related to collections.
 repository: https://github.com/dart-lang/core/tree/main/pkgs/collection
--- a/pkgs/collection/test/extensions_test.dart
+++ b/pkgs/collection/test/extensions_test.dart
@@ -1122,6 +1122,87 @@ void main() {
         });
       });
     });
+    group('of MapEntry', () {
+      group('.whereKey', () {
+        test('empty', () {
+          expect(<String, int>{}.entries.whereKey(unreachable), isEmpty);
+        });
+        test('single', () {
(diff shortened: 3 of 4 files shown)


pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -914,6 +914,33 @@ extension IterableIterableExtension<T> on Iterable<Iterable<T>> {
       };
 }
 
+/// Extension on iterables of [MapEntry].
+///
+/// An [Iterable<MapEntry>] is obtained using [Map.entries], these extensions
+/// make it easy to work on a [Map] as a list of pairs.
+extension IterableMapEntryExtension<K, V> on Iterable<MapEntry<K, V>> {
+  /// Creates a new lazy [Iterable] with all elements whose [MapEntry.key]
+  /// satisfy the predicate [test].
```
lrhn, line 923: Long first line. I'd also use a noun-phrased name to match the predicate named function.

Consider:
```dart
  /// The elements whose [MapEntry.key] values satisfy [test].
  ///
  /// The resulting iterable is lazily computing its elements 
  /// based on the elements this iterable.
```
Or something.

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -914,6 +914,33 @@ extension IterableIterableExtension<T> on Iterable<Iterable<T>> {
...
+  Iterable<MapEntry<K, V>> whereKey(bool Function(K) test) =>
+      where((e) => test(e.key));
+
+  /// Creates a new lazy [Iterable] with all elements whose [MapEntry.value]
+  /// satisfy the predicate [test].
+  Iterable<MapEntry<K, V>> whereValue(bool Function(V) test) =>
+      where((e) => test(e.value));
+
+  /// Create an new lazy [Iterable] with [MapEntry.key] from all elements.
```
lrhn, line 932: Maybe say:
```dart
  ///
  /// Do not use this getter as `map.entries.keys`, just use `map.keys` directly.
```

lrhn, line 932: Drop the "Create". (Should be "Creates" if it was there).

Just 
```dart
  /// A new lazy [Iterable] of the [MapEntry.key]s of these entries.
```   
It's cleaner.

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -914,6 +914,33 @@ extension IterableIterableExtension<T> on Iterable<Iterable<T>> {
...
+  Iterable<K> get keys => map((e) => e.key);
+
+  /// Create an new lazy [Iterable] with [MapEntry.value] from all elements.
```
lrhn, line 935: Ditto.

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -914,6 +914,33 @@ extension IterableIterableExtension<T> on Iterable<Iterable<T>> {
...
+  Iterable<V> get values => map((e) => e.value);
+
+  /// Create a [Map] from all elements.
```
lrhn, line 938: And
```dart
  /// Creates a new [Map<K, V>] with these entries.
  ///
  /// This is ...
```

pkgs/collection/test/extensions_test.dart
```diff
@@ -1122,6 +1122,87 @@ void main() {
         });
       });
     });
+    group('of MapEntry', () {
+      group('.whereKey', () {
+        test('empty', () {
+          expect(<String, int>{}.entries.whereKey(unreachable), isEmpty);
+        });
+        test('single', () {
+          expect({'a': 1}.entries.whereKey((k) => k == 'a').toMap(), {'a': 1});
```
lrhn, line 1131: Don't use `toMap` in the test of `whereKey`. If there is a bug in `toMap`, it might hide a bug in `whereKey`.

Check against `[MapEntry('a', 1)]` instead.

I'd even say to not use `map.entries` as the source. Just have something like:
```dart
Iterable<MapEntry<K, V>> asIterable<K, V>(Iterable<MapEntry<K, V>> entries) => entries;
// ...
  expect(asIterable([MapEntry('a', 1)]).whereKey((k) => k == 'a'), [MapEntry('a', 1)]);
```
The `asIterable` is only there to change the static type away from `List<MapEntry>`, in case that matters. (It shouldn't, but this is testing, not assuming things work.)

pkgs/collection/test/extensions_test.dart
```diff
@@ -1122,6 +1122,87 @@ void main() {
...
+          );
+          expect(
+            {'a': 1, 'b': 2}.entries.whereKey((k) => k == 'b').toMap(),
+            {'b': 2},
+          );
+          expect(
+            {'a': 1, 'b': 2}.entries.whereKey((k) => k != 'c').toMap(),
+            {'a': 1, 'b': 2},
+          );
+        });
```
lrhn, line 1147: Consider checking an iterable of entries which have duplicate keys. Fx.
```dart
expect([MapEntry('a': 1), MapEntry('b': 2), MapEntry('a': 3)].whereKey((k) => k == 'a'), 
   [MapEntry('a': 1), MapEntry('a': 3)]);
```

pkgs/collection/test/extensions_test.dart
```diff
@@ -1122,6 +1122,87 @@ void main() {
...
+      group('.keys', () {
+        test('empty', () {
+          expect(<String, int>{}.entries.keys, isEmpty);
+        });
+        test('single', () {
+          expect({'a': 1}.entries.keys, ['a']);
+        });
+        test('multiple', () {
+          expect({'a': 1, 'b': 2}.entries.keys, ['a', 'b']);
+        });
```
lrhn, line 1181: Repeated keys here too.
```dart
  expect([MapEntry('a', 1), MapEntry('b': 2), MapEntry('a': 3)].keys, ['a', 'b', 'a']);
```

pkgs/collection/test/extensions_test.dart
```diff
@@ -1122,6 +1122,87 @@ void main() {
...
+      group('.toMap', () {
+        test('empty', () {
+          expect(<String, int>{}.entries.toMap(), <String, int>{});
+        });
+        test('single', () {
+          expect({'a': 1}.entries.toMap(), {'a': 1});
+        });
+        test('multiple', () {
+          expect({'a': 1, 'b': 2}.entries.toMap(), {'a': 1, 'b': 2});
+        });
```
lrhn, line 1203: Definitely have cases with repeated keys here:
```dart
  expect([MapEntry('a', 1), MapEntry('b', 2), MapEntry('a', 3)].toMap(), {'a': 3, 'b': 2});
```
(You don't have to document the behavior of multiple occurrences of a key, the `toMap` refers to `Map.fromEntries` which does document it.)

lrhn (approved):
I like it.
I'm not awfully happy about people working on `MapEntry`s in general, but this seems to be withing  reason (it's not just treating them as arbitrary pairs).

pkgs/collection/lib/src/iterable_extensions.dart
jonasfj: I actually took the documentation string from `.where` which says:
```dart
>   /// Creates a new lazy [Iterable] with all elements that satisfy the
>   /// predicate [test].
>   ///
>   /// ...
```

I don't really care, the behavior is implied by the name, and the implementation is absolutely trivial to follow :D

pkgs/collection/lib/src/iterable_extensions.dart
jonasfj: Weird syntax highlighting convinced me that wouldn't work, but it works fine in dartdoc :rocket: 

image

Filed https://github.com/dart-lang/dart-syntax-highlight/issues/77 for the syntax issue.

pkgs/collection/test/extensions_test.dart
jonasfj: Done, I don't mind, but we are effectively testing the behavior of the Dart SDK.

So I'm not really sure this does much other than make the tests harder to read :rofl:

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -914,6 +914,33 @@ extension IterableIterableExtension<T> on Iterable<Iterable<T>> {
...
+extension IterableMapEntryExtension<K, V> on Iterable<MapEntry<K, V>> {
+  /// Creates a new lazy [Iterable] with all elements whose [MapEntry.key]
+  /// satisfy the predicate [test].
+  Iterable<MapEntry<K, V>> whereKey(bool Function(K) test) =>
+      where((e) => test(e.key));
+
+  /// Creates a new lazy [Iterable] with all elements whose [MapEntry.value]
+  /// satisfy the predicate [test].
+  Iterable<MapEntry<K, V>> whereValue(bool Function(V) test) =>
+      where((e) => test(e.value));
```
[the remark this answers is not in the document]
jonasfj: Yes, I feel the same with `.map`.

But I wouldn't make `where({bool Function(K)? key, bool Function(V)? value})`.
I would make: `where(bool Function(K, V) test)`, ignoring a parameter in a closure is not that bad.

Maybe, we should consider adding:
 * `wherePair(bool Function(K, V) test)`
 * `mapPair<T>(T Function(K, V) toElement)`
We could consider adding such extension methods to:
 * `Iterable<MapEntry<K, V>>`
 * `Iterable<(A, B)>`

pkgs/collection/lib/src/iterable_extensions.dart
lrhn: Yeah, that's too long too. But OK,  keep it here, it's not *that* too long.

But consider changing "elements" to "entries", since all elements of this iterable are (map-)entries, not just elements. (I'd refer to the elements of an `Iterable<num>` as "the numbers" too.)

pkgs/collection/lib/src/iterable_extensions.dart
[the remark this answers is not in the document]
lrhn: Maybe `whereEntry`/`mapEntry`. (Don't say "pair", it's not &mdash; just &mdash; a pair! It's a *key-and-value*.)

lrhn (commented):
LGTM.

pkgs/collection/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: collection
-version: 1.19.1
+version: 1.19.1-wip
```
lrhn, line 2: Should be 1.20.0, unless 1.19.0 hasn't been released yet.
(Adds new feature &rArr; minor version increment.)

PR #718, opened 2024-11-05
Use `values` iterator  and `update` in map equality/hash

jakemac53:
Initially I tried using `entries` instead of a `values` iterator, but that was surprisingly slow. This appears to be a good middle ground (see history of this comment for old results).

Hashing is about the same, but much faster for maps which don't have O(1) lookup. Equality is much faster.

## Benchmarks before:

DeepCollectionQualityUnordered.equals(RunTime): 7427.29020979021 us.
DeepCollectionQualityUnordered.hash(RunTime): 217.8173707406213 us.
DeepCollectionQuality.equals(RunTime): 2653.23875 us.
DeepCollectionQuality.hash(RunTime): 178.1674653887114 us.

## Benchmarks after:

DeepCollectionQualityUnordered.equals(RunTime): 4435.374 us.
DeepCollectionQualityUnordered.hash(RunTime): 212.8631545473818 us.
DeepCollectionQuality.equals(RunTime): 1989.1746626686656 us.
DeepCollectionQuality.hash(RunTime): 178.3396697902722 us.

### DIFF
--- a/pkgs/collection/lib/src/equality.dart
+++ b/pkgs/collection/lib/src/equality.dart
@@ -329,14 +329,14 @@ class MapEquality<K, V> implements Equality<Map<K, V>> {
     for (var key in map1.keys) {
       var value = (values1..moveNext()).current;
       var entry = _MapEntry(this, key, value);
-      equalElementCounts.update(entry, (i) => i + 1, ifAbsent: () => 1);
+      equalElementCounts.update(entry, _addOne, ifAbsent: _one);
     }
     final values2 = map2.values.iterator;
     for (var key in map2.keys) {
       var value = (values2..moveNext()).current;
       var entry = _MapEntry(this, key, value);
-      var count =
-          equalElementCounts.update(entry, (i) => i - 1, ifAbsent: () => -1);
+      var count = equalElementCounts.update(entry, _subtractOne,
+          ifAbsent: _negativeOne);
       if (count < 0) return false;
     }
     return true;
@@ -494,3 +494,8 @@ class CaseInsensitiveEquality implements Equality<String> {
   @override
   bool isValidKey(Object? object) => object is String;
 }
+
+int _addOne(int i) => i + 1;
+int _subtractOne(int i) => i - 1;
+int _one() => 1;
+int _negativeOne() => -1;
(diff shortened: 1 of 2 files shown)


jakemac53:
If we don't want to land the `entries` changes, I will end up having to fork this implementation. But, that is probably OK, just a bit unfortunate.

jakemac53:
cc @davidmorgan

pkgs/collection/lib/src/equality.dart
```diff
@@ -325,16 +325,19 @@ class MapEquality<K, V> implements Equality<Map<K, V>> {
...
     if (length != map2.length) return false;
     Map<_MapEntry, int> equalElementCounts = HashMap();
+    var values1 = map1.values.iterator;
     for (var key in map1.keys) {
-      var entry = _MapEntry(this, key, map1[key]);
-      var count = equalElementCounts[entry] ?? 0;
-      equalElementCounts[entry] = count + 1;
+      var value = (values1..moveNext()).current;
+      var entry = _MapEntry(this, key, value);
+      equalElementCounts.update(entry, (i) => i + 1, ifAbsent: () => 1);
```
natebosch, line 332: Can we define these callbacks statically instead of creating them on each iteration?

pkgs/collection/lib/src/equality.dart
```diff
@@ -325,16 +325,19 @@ class MapEquality<K, V> implements Equality<Map<K, V>> {
...
+    final values2 = map2.values.iterator;
     for (var key in map2.keys) {
-      var entry = _MapEntry(this, key, map2[key]);
-      var count = equalElementCounts[entry];
-      if (count == null || count == 0) return false;
-      equalElementCounts[entry] = count - 1;
+      var value = (values2..moveNext()).current;
+      var entry = _MapEntry(this, key, value);
+      var count =
+          equalElementCounts.update(entry, (i) => i - 1, ifAbsent: () => -1);
```
natebosch, line 339: Can we define these callbacks statically?

natebosch (approved):

pkgs/collection/lib/src/equality.dart
jakemac53: Done, it doesn't really make a difference on the VM but it might on other platforms (I think its optimized well though).

pkgs/collection/lib/src/equality.dart
jakemac53: Done

PR #719, opened 2024-11-07
Implement full Unicode 16.0.0 extended grapheme breaking.

lrhn:
Includes rule GB9c (Indic Conjunt Break rule).

This change has a significant cost in size since the information needed per character no longer fits in 4 bits. The base table is therefore twice as big (one byte per entry rather than half of that).

The number of states in the state automatons have also increased slightly, but in comparison that's a negligible change.

Tests have been made more thorough, testing not only the Unicode Consortium provided tests, but also variants of those with representative characters for each category of character that either in or not-in the BMP, to test that surrogate pair decoding works correctly.

Test also check that the created automatons are minimal, in that no state is unreachable and no two states are indistinguishable.


lrhn:
Health check is wrong. The changelog is correct since the version wasn't changed, and the existing changelog didn't list missing part that is now implemented.

pkgs/characters/tool/src/graph.dart
```diff
@@ -0,0 +1,82 @@
...
+      var weightFrom2 = weight(vertex2, i);
+      setWeight(vertex1, i, weightFrom2);
+      setWeight(vertex2, i, weightFrom1);
+    }
+    assert(weight(vertex1, vertex1) == 0);
+    assert(weight(vertex2, vertex2) == 0);
+  }
+
+  /// Creates a new graph without the last vertex (or last [count] vertices).
+  Graph removeLastVertex([int count = 1]) {
```
natebosch, line 70: I cannot find where this is used. Can it be removed?

pkgs/characters/tool/src/graph.dart
```diff
@@ -0,0 +1,82 @@
...
+      i += step;
+      var next = path[i];
+      weight += this.weight(cursor, next);
+      cursor = next;
+    }
+    return weight;
+  }
+
+  /// Swaps the positions of [vertex1] and [vertex2], updating the weight table.
+  void swap(int vertex1, int vertex2) {
```
natebosch, line 49: I cannot find where this is used. Can it be removed?

pkgs/characters/lib/src/grapheme_clusters/breaks.dart
```diff
@@ -330,52 +467,250 @@ int previousBreak(String text, int start, int end, int index) {
 
 /// The next break no earlier than [index] in `string.substring(start, end)`.
 ///
-/// The index need not be at a grapheme cluster boundary.
+/// Uniquely to this function, the index need not be at a grapheme cluster
+/// boundary. That means there may be need for look-behind to find a character
+/// where the exact state is known.
 int nextBreak(String text, int start, int end, int index) {
```
natebosch, line 473: [nit] This method is going from <50 lines to >150 lines. Consider whether any pieces  can be pulled out into composed methods.

natebosch (commented):
Out of these comments my primary concern is the `return` statements in loops in tests.

Should some of those be `continue` instead of `return`?

pkgs/characters/tool/src/graph.dart
lrhn: Same.

pkgs/characters/lib/src/grapheme_clusters/breaks.dart
lrhn: I'm considering whether it should be a state machine. It probably can be. Maybe part of the forwards automaton, with more states above the normal states, since that's the normal states it should end up in after figuring out what it needs from the look-behind.

pkgs/characters/test/breaks_test.dart
```diff
@@ -0,0 +1,366 @@
...
+                }
+              }
+              nextEq.equate(state1, state2);
+            }
+          }
+        }
+        var nextEqClasses = nextEq.classes;
+        if (nextEqClasses.length == eqClasses.length) break;
+        if (nextEqClasses.length == states.length) {
+          print("Backwards states distinguishable in $r steps");
```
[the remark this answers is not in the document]
lrhn: This test is a sanity check. It's assumed that all the breaking-tests succeed.
These tests just test that the tables are not too big.

The print is just to give me a little information (I found bugs in the test because the value looked wrong).

pkgs/characters/lib/src/grapheme_clusters/breaks.dart
lrhn: And I've done that. More tables, more better!

pkgs/characters/test/breaks_test.dart
```diff
@@ -0,0 +1,366 @@
...
+            if (unreachableStates.remove(newState)) {
+              visit(stateExtend);
+            }
+          } else {
+            visit(newState);
+          }
+        }
+        if (unreachableStates.isEmpty) {
+          print("Backward states reachable in $step steps");
+          return;
```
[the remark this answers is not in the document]
lrhn: It's a succes branch. It should get hit, but we shouldn't assume it. This test failing can be a canary for something else being wrong (and possibly a good hint at what).

pkgs/characters/test/breaks_test.dart
```diff
@@ -0,0 +1,366 @@
...
+        if (nextEqClasses.length == states.length) {
+          print("Backwards states distinguishable in $r steps");
+          return; // All needed.
+        }
+        eqClasses = nextEqClasses;
+      }
+      expect(
+        eqClasses.where((l) => l.length != 1).toList(),
+        isEmpty,
+      );
```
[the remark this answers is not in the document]
lrhn: `everyElement(hasLength(1))`? (Can't find an `every`.)

pkgs/characters/test/breaks_test.dart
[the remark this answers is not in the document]
natebosch: if the return is expected the other branch should call `fail` instead of `expect`

pkgs/characters/test/breaks_test.dart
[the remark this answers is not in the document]
natebosch: Ah yes, it's called `everyElement` in matcher.

lrhn:
> Out of these comments my primary concern is the `return` statements in loops in tests.
> 
> Should some of those be `continue` instead of `return`?

They could be `break`, and now they are.
(They should break to an `expect` that will not be able to fail, but that still means every test run goes through that `expect`, one way or another.)

lrhn:
Think I'm done now. Much cleaned up (and nicer, IMO).

natebosch (approved):

PR #731, opened 2024-12-06
add `ascending` to IterableExtension.sortedBy

arthurbcd:
`sortedBy` is one of the best extensions of this package.

It allows us to easily sort, but we can't set if the order comparaison is ascending/descending, it always defaults to "natural ordering".

This PR simplifies this process so we can simply decide it based on the added `ascending` parameter in the `sortedBy` extension.

```dart
  List<T> sortedBy<K extends Comparable<K>>(
    K Function(T element) keyOf, {
    bool ascending = true,
  })
```

This is useful, because without it, we would have to do additional operations bringing either more boilerplate or additional computations. By adding this named boolean we can make it straightforward.

For many developers, especially those who might not be as familiar with comparators and more advanced functional programming concepts, having a straightforward boolean parameter makes the API more accessible and user-friendly.

The boolean `ascending` is a natural state. It’s a common need to toggle sort order, and doing so with a single extension method and a simple parameter makes the code more readable and maintainable. Adopting this familiar pattern in `sortedBy` makes it feel more intuitive and consistent, making the intent of the code clearer.

Flutter itself uses this pattern for sorting, as we can see in `flutter/lib/src/material/data_table.dart`:

```dart
/// Signature for [DataColumn.onSort] callback.
typedef DataColumnSortCallback = void Function(int columnIndex, bool ascending);
```

I find this extension incredibly useful and have always relied on a version of it. I dedicated time to updating it for the community because I believe it addresses a fundamental and widely shared need across projects.

Thank you for taking the time to review this PR—I hope it proves valuable!

Added tests ✅
Retrocompatible ✅
Not a breaking change ✅

---

<details>

</details>

### DIFF
--- a/pkgs/collection/lib/src/iterable_extensions.dart
+++ b/pkgs/collection/lib/src/iterable_extensions.dart
@@ -6,7 +6,6 @@ import 'dart:math' show Random;
 
 import 'algorithms.dart';
 import 'functions.dart' as functions;
-import 'utils.dart';
 
 /// Extensions that apply to all iterables.
 ///
@@ -71,10 +70,15 @@ extension IterableExtension<T> on Iterable<T> {
   /// Creates a sorted list of the elements of the iterable.
   ///
   /// The elements are ordered by the natural ordering of the
-  /// property [keyOf] of the element.
-  List<T> sortedBy<K extends Comparable<K>>(K Function(T element) keyOf) {
+  /// [keyOf] property when [ascending] is `true` or reverse ordering
+  /// when [ascending] is `false`.
+  List<T> sortedBy<K extends Comparable<K>>(
+    K Function(T element) keyOf, {
+    bool ascending = true,
+  }) {
+    int compare(K a, K b) => ascending ? a.compareTo(b) : b.compareTo(a);
     var elements = [...this];
-    mergeSortBy<T, K>(elements, keyOf, compareComparable);
+    mergeSortBy<T, K>(elements, keyOf, compare);
     return elements;
   }
 
--- a/pkgs/collection/test/extensions_test.dart
+++ b/pkgs/collection/test/extensions_test.dart
@@ -64,6 +64,16 @@ void main() {
         test('multiple', () {
           expect(iterable(<int>[3, 20, 100]).sortedBy(toString), [100, 20, 3]);
         });
+        test('multiple ascending', () {
+          expect(
+              iterable(<int>[3, 20, 100]).sortedBy(toString, ascending: true),
+              [100, 20, 3]);
+        });
+        test('multiple descending', () {
+          expect(
+              iterable(<int>[3, 20, 100]).sortedBy(toString, ascending: false),
+              [3, 20, 100]);
+        });
       });
       group('.sortedByCompare', () {
         test('empty', () {


natebosch (commented):
My initial inclination is to not add this - I don't think we have a similar argument in other sorting methods, we instead expect callers to arrange an appropriate compare callback.

Here I think it's better to ask users to call `sortedByCompare` for this use case.

cc @lrhn WDYT?

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -71,10 +70,15 @@ extension IterableExtension<T> on Iterable<T> {
...
   /// The elements are ordered by the natural ordering of the
-  /// property [keyOf] of the element.
-  List<T> sortedBy<K extends Comparable<K>>(K Function(T element) keyOf) {
+  /// [keyOf] property when [ascending] is `true` or reverse ordering
+  /// when [ascending] is `false`.
+  List<T> sortedBy<K extends Comparable<K>>(
+    K Function(T element) keyOf, {
+    bool ascending = true,
+  }) {
+    int compare(K a, K b) => ascending ? a.compareTo(b) : b.compareTo(a);
```
natebosch, line 79: At the risk of premature optimization, I think we should keep this as a static tear-off instead of a new lambda on every use. If we decide to go forward with this change I think something like

```dart
final compare = ascending ? compareComparable : compareComparableDescending;

int compareComparableDescending<T extends Comparable<T>>(T a, T b) => b.compareTo(a);
```

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -71,10 +70,15 @@ extension IterableExtension<T> on Iterable<T> {
   /// Creates a sorted list of the elements of the iterable.
   ///
   /// The elements are ordered by the natural ordering of the
-  /// property [keyOf] of the element.
-  List<T> sortedBy<K extends Comparable<K>>(K Function(T element) keyOf) {
+  /// [keyOf] property when [ascending] is `true` or reverse ordering
+  /// when [ascending] is `false`.
+  List<T> sortedBy<K extends Comparable<K>>(
+    K Function(T element) keyOf, {
+    bool ascending = true,
```
lrhn, line 77: I think I would prefer to name it `descending`, so the default is `false`.
Just because it's usually best to have "nothing" mean `false`.
Then it feels like you are *opting in* to the non-default behavior. Using `ascending: false` feels ... conflicted. You are opting *out* of the default behavior, but that doesn't directly say what it opts in to.

pkgs/collection/lib/src/iterable_extensions.dart
lrhn: If it's typed, it's a new closure on every use anyway, because generic instantiation also causes a new closure.
That said: I don't want to do the *branch*  on every comparison, so something like what @natebosch says is better.

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -71,10 +70,15 @@ extension IterableExtension<T> on Iterable<T> {
   /// Creates a sorted list of the elements of the iterable.
   ///
   /// The elements are ordered by the natural ordering of the
-  /// property [keyOf] of the element.
-  List<T> sortedBy<K extends Comparable<K>>(K Function(T element) keyOf) {
+  /// [keyOf] property when [ascending] is `true` or reverse ordering
```
lrhn, line 73: The `sorted` function should get the same behavior. Really, every `sort` function should behave the same way if at all possible.
(It's probably impossible because there is *some* `sort([int Function(T, T) compare])` somewhere which prevents a named parameter.)

lrhn (approved):
I think it's an OK feature.
I would reverse the parameter name (to make he default value `false`) and give every sorting function the same treatment (if possible).

mosuem:
@arthurbcd do you still intend to land this PR? If so, this is a friendly ping!

pkgs/collection/lib/src/iterable_extensions.dart
arthurbcd: I understand the concern about double negatives with `ascending: false`. However, I'd lean towards `ascending: true` in this case because:

1. Flutter's `DataTable` already uses `sortAscending: true` as the default, and `DataColumnSortCallback` uses `bool ascending` in its signature
2. In sorting contexts across Flutter/Dart, I feel developers would expect the `ascending` parameter

That said, I recognize the clarity benefit of `descending: true` for opt-in behavior. If you still prefer `descending: false` after considering these points, I'm happy to update it accordingly. :)

pkgs/collection/lib/src/iterable_extensions.dart
arthurbcd: Updated it, removing the _branch_:
```dart
    final compare =
        ascending ? (K a, K b) => a.compareTo(b) : (K a, K b) => b.compareTo(a);
```

arthurbcd:
@lrhn @mosuem @brianquinlan Do I still need to do anything?

arthurbcd:
@natebosch you can handle and land it :) thanks for the discussion !

pkgs/collection/lib/src/iterable_extensions.dart
arthurbcd: That’s a fair point. My previous references align more with state variables or callbacks, but for defining intent, I like your reasoning about 'stating what is'. So, let's do it this way :)

And thanks for the discussion!

PR #734, opened 2024-12-18
Adds `[Heap]PriorityQueue.of` constructor.

### LINKED ISSUE #732: Implement heapify in `HeapPriorityQueue.addAll`
Current `HeapPriorityQueue` calls `add` for each element in the given Iterable yielding a time complexity of $O(n * logn)$, but if the heapify algorithm is used instead the time complexity can be reduced to $O(n)$. This has been done in other languages such as Go for example: https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/container/heap/heap.go;l=41

lrhn:
Introduces efficient (linear-number of comparisons) "heapify" algorithm for converting an unsorted list to a heap-sorted list, using it for the `of` constructor, and after a large `addAll` operation, when it's presumed faster than just bubbling down all the new elements.

Also rewrites `HeapPriorityQueue` to use a growable list as backing array, instead of implementing the same thing using the double-when-full algorithm, and still having to deal with nullable cells. The platform growable list implementation is assumed to efficiently avoid some of those `null` checks.

Closes #732

### DIFF
--- a/pkgs/collection/lib/src/priority_queue.dart
+++ b/pkgs/collection/lib/src/priority_queue.dart
@@ -22,9 +22,9 @@ import 'utils.dart';
 /// always give equal objects the same priority,
 /// otherwise [contains] or [remove] might not work correctly.
 abstract class PriorityQueue<E> {
-  /// Creates an empty [PriorityQueue].
+  /// Creates an empty priority queue.
   ///
-  /// The created [PriorityQueue] is a plain [HeapPriorityQueue].
+  /// The created `PriorityQueue` is a plain [HeapPriorityQueue].
   ///
   /// The [comparison] is a [Comparator] used to compare the priority of
   /// elements. An element that compares as less than another element has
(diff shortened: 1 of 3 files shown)


pkgs/collection/lib/src/priority_queue.dart
```diff
@@ -36,6 +36,19 @@ abstract class PriorityQueue<E> {
...
+  /// Create a new priority queue.
+  ///
+  /// The [comparison] is a [Comparator] used to compare the priority of
+  /// elements. An element that compares as less than another element has
+  /// a higher priority.
+  ///
+  /// If [comparison] is omitted, it defaults to [Comparable.compare]. If this
+  /// is the case, `E` must implement [Comparable], and this is checked at
+  /// runtime for every comparison.
+  factory PriorityQueue.of(
```
kevmoo, line 48: What's the right way to easily support `Comparable<T>`

`factory PriorityQueue.ofComparable<T extends Comparable<T>>(Iterable<T>)`...something?

lrhn: That's an option, or just use `Comparable.compare` explicitly.

lrhn: (But I have not current plans to add any more "optional only if type argument is comparable/nullable" parameters, so for now you'll just have to pass `Comparable.compare` as the comparison.)

lrhn:
Ping @natebosch

natebosch (approved):

devoncarew:
We're seeing a google3 roll failure, I believe from this change:

```
[.../async/lib/src/priority_stream_iterator.dart:110]
'_StablePriorityQueue.removeAll' ('Iterable<T> Function()') isn't a valid override of 'HeapPriorityQueue.removeAll' ('List<T> Function()'). #invalid_override

  Iterable<T> removeAll() {
              ^^^^^^^^^
```

PR #735, opened 2024-12-20
Update package:collection for the new `strict_top_level_inference` lint

devoncarew:
- update package:collection readme to get analysis to run

---

<details>

</details>

### DIFF
--- a/pkgs/collection/lib/src/unmodifiable_wrappers.dart
+++ b/pkgs/collection/lib/src/unmodifiable_wrappers.dart
@@ -193,12 +193,4 @@ abstract mixin class UnmodifiableMapMixin<K, V> implements Map<K, V> {
   /// operations that change the map are disallowed.
   @override
   void clear() => _throw();
-
-  /// Throws an [UnsupportedError];
-  /// operations that change the map are disallowed.
-  set first(_) => _throw();
-
-  /// Throws an [UnsupportedError];
-  /// operations that change the map are disallowed.
-  set last(_) => _throw();
 }
--- a/pkgs/collection/lib/src/unmodifiable_wrappers.dart
+++ b/pkgs/collection/lib/src/unmodifiable_wrappers.dart
@@ -193,4 +193,12 @@ abstract mixin class UnmodifiableMapMixin<K, V> implements Map<K, V> {
   /// operations that change the map are disallowed.
   @override
   void clear() => _throw();
+
+  /// Throws an [UnsupportedError].
+  @Deprecated('This will be removed at a later date.')
+  set first(dynamic _) => _throw();
+
+  /// Throws an [UnsupportedError].
+  @Deprecated('This will be removed at a later date.')
+  set last(dynamic _) => _throw();
 }
--- a/pkgs/collection/lib/src/unmodifiable_wrappers.dart
+++ b/pkgs/collection/lib/src/unmodifiable_wrappers.dart
@@ -195,10 +195,10 @@ abstract mixin class UnmodifiableMapMixin<K, V> implements Map<K, V> {
   void clear() => _throw();
 
   /// Throws an [UnsupportedError].
-  @Deprecated('This will be removed at a later date.')
-  set first(dynamic _) => _throw();
+  @deprecated
+  set first(Object? _) => _throw();
 
   /// Throws an [UnsupportedError].
-  @Deprecated('This will be removed at a later date.')
-  set last(dynamic _) => _throw();
+  @deprecated
+  set last(Object? _) => _throw();
 }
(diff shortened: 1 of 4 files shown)


devoncarew:
From the new `strict_top_level_inference` lint, we see two sets of issues here:

```
   info - lib/src/unmodifiable_wrappers.dart:199:7 - Missing type annotation. Try adding a type annotation. - strict_top_level_inference
   info - lib/src/unmodifiable_wrappers.dart:203:7 - Missing type annotation. Try adding a type annotation. - strict_top_level_inference
```
That's from https://github.com/dart-lang/core/blob/main/pkgs/collection/lib/src/unmodifiable_wrappers.dart#L199:

```dart
  /// Throws an [UnsupportedError];
  /// operations that change the map are disallowed.
  set first(_) => _throw();

  /// Throws an [UnsupportedError];
  /// operations that change the map are disallowed.
  set last(_) => _throw();
```

Here, I don't know what purpose `first` and `last` are for; they're not overrides from `Map`.

The next set of issues:

```
   info - test/extensions_test.dart:2228:20 - Missing type annotation. Try adding a type annotation. - strict_top_level_inference
   info - test/extensions_test.dart:2228:23 - Missing type annotation. Try adding a type annotation. - strict_top_level_inference
   info - test/extensions_test.dart:2228:27 - Missing type annotation. Try adding a type annotation. - strict_top_level_inference
```

are from https://github.com/dart-lang/core/blob/main/pkgs/collection/test/extensions_test.dart#L2228:

```dart
Never unreachable([_, __, ___]) => fail('Unreachable');
```

Perhaps we should not include wildcard variables in the lint?

cc @lrhn, @natebosch, @srawlins

lrhn:
The setters should just be removed from the `Map` wrapper, they're mistakes.

Other than that, we could exclude wildcards, but since the lint should only trigger if you don't inherit a type, I think the method should actually be explicit about its types.

pkgs/collection/lib/src/unmodifiable_wrappers.dart
```diff
@@ -193,12 +193,4 @@ abstract mixin class UnmodifiableMapMixin<K, V> implements Map<K, V> {
   /// operations that change the map are disallowed.
   @override
   void clear() => _throw();
-
-  /// Throws an [UnsupportedError];
-  /// operations that change the map are disallowed.
-  set first(_) => _throw();
```
jakemac53, line 199: Technically this is breaking to remove these?

We might want to just deprecate them 🤷‍♂️ . But should at least verify we don't know of any usages (I would expect none though since nobody really uses this type directly in APIs).

devoncarew: I'm happy to deprecate first and check for any real usage asynchronously.

pkgs/collection/lib/src/unmodifiable_wrappers.dart
```diff
@@ -193,4 +193,12 @@ abstract mixin class UnmodifiableMapMixin<K, V> implements Map<K, V> {
   /// operations that change the map are disallowed.
   @override
   void clear() => _throw();
+
+  /// Throws an [UnsupportedError].
+  @Deprecated('This will be removed at a later date.')
```
lrhn, line 198: Can we just use `@deprecated` if the message isn't saying anything  anyway.

pkgs/collection/lib/src/unmodifiable_wrappers.dart
```diff
@@ -193,4 +193,12 @@ abstract mixin class UnmodifiableMapMixin<K, V> implements Map<K, V> {
...
+  set first(dynamic _) => _throw();
```
lrhn, line 199: Could we make the parameter type be `Object?` or `void`?

It should make no difference in a contravariant position, and I always get wary when I see `dynamic`.
(I think it makes not difference. With enum shorthands, I guess `Object?` will allow `.hash(...)`, so maybe just use `void`.)

pkgs/collection/lib/src/unmodifiable_wrappers.dart
lrhn: If we do a proper `3.0` release soon-ish,  they can go away there.

lrhn (approved):

pkgs/collection/lib/src/unmodifiable_wrappers.dart
devoncarew: done

pkgs/collection/lib/src/unmodifiable_wrappers.dart
devoncarew: I switched to `Object?` to avoid the use of dynamic.

jakemac53 (approved):

jakemac53:
> If we do a proper 3.0 release soon-ish, they can go away there.

Note that this package is pinned by flutter so this is probably never going to be justified.... it will cause a very large amount of ecosystem churn and pain. Unless of course flutter stops pinning packages :)

PR #850, opened 2025-01-13
[os_detect] move to pkg:web, prepare publish

kevmoo:


devoncarew (approved):
lgtm, but lets be sure to run this through the regular pipeline (sdk + google3 roll) before publishing.

PR #852, opened 2025-01-17
Remove sorting of allowedHelp maps

### LINKED ISSUE #845: [args] Inconsistent order between `allowed` and `allowedHelp`
Hello,

Not sure if that was ever discussed before, but `allowed` respects user defined order of options and `allowedHelp` for some reason [sorts them](https://github.com/dart-lang/core/blob/main/pkgs/args/lib/src/usage.dart#L94). This sorting behavior is really inconvenient, and, for example, for numbers results in: 

```text
    --rule-version=<all>                           Rule version to include (all by default).

          [1.0.0]                                  Included rules added in 1.0.0.
          [1.1.0]                                  Included rules added in 1.1.0.
          [1.10.0]                                 Included rules added in 1.10.0.
          [1.11.0]                                 Included rules added in 1.11.0.
          [1.2.0]                                  Included rules added in 1.2.0.
```

which is simply confusing for the end user.

Is there an option to make this behaviour configurable or remove it completely? Considering the inconsistency, it seems like giving the user full control over the order of items makes more sense.

natebosch:
Closes #845

It is idiomatic to treat the key order of a Dart map as meaningful
given that map literals and default Map type preserve key insertion
order. It is more useful to allow the caller to decide this order than
to mandate an alpha sorting by key. Callers which need this order can
construct the map appropriately, and callers which prefer a different
order now have the capability.

Remove the additional list copy and iterate the map keys directly.

Releasing as a non-breaking change since specific usage output is
considered an implementation detail. This is expected to impact some CI
statuf for packages with tests hardcoding a strict dependency on the
output.

No additional tests are necessary since updating the order in existing
tests demonstrates the same behavior as adding a non-sorting specific
test.

### DIFF
--- a/pkgs/args/lib/src/usage.dart
+++ b/pkgs/args/lib/src/usage.dart
@@ -90,8 +90,7 @@ class _Usage {
     if (option.help != null) _write(2, option.help!);
 
     if (option.allowedHelp != null) {
-      var allowedNames = option.allowedHelp!.keys.toList();
-      allowedNames.sort();
+      var allowedNames = [...option.allowedHelp!.keys];
       _newline();
       for (var name in allowedNames) {
         _write(1, _allowedTitle(option, name));
--- a/pkgs/args/lib/src/usage.dart
+++ b/pkgs/args/lib/src/usage.dart
@@ -90,9 +90,8 @@ class _Usage {
     if (option.help != null) _write(2, option.help!);
 
     if (option.allowedHelp != null) {
-      var allowedNames = [...option.allowedHelp!.keys];
       _newline();
-      for (var name in allowedNames) {
+      for (var name in option.allowedHelp!.keys) {
         _write(1, _allowedTitle(option, name));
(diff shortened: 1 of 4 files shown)


pkgs/args/lib/src/usage.dart
```diff
@@ -90,10 +90,8 @@ class _Usage {
     if (option.help != null) _write(2, option.help!);
 
     if (option.allowedHelp != null) {
-      var allowedNames = option.allowedHelp!.keys.toList();
-      allowedNames.sort();
       _newline();
-      for (var name in allowedNames) {
+      for (var name in option.allowedHelp!.keys) {
```
jakemac53, line 94: nit: consider an if case to extract this as a variable and avoid the `!`

jakemac53 (approved):

pkgs/args/lib/src/usage.dart
natebosch: Done, did a couple nearby ones as well.

PR #855, opened 2025-01-24
[fix,async] exception in async _CancelOnErrorSubscriptionWrapper.onError

PROGrand:
Patch fixes exception in async (template cast seems to be changed in dart):

```
Unhandled exception:
type '(Object, StackTrace) => void' is not a subtype of type '(dynamic) => dynamic' in type cast
#0      _CancelOnErrorSubscriptionWrapper.onError.<anonymous closure>.<anonymous closure> (package:async/src/subscription_stream.dart:82:24)
#1      _RootZone.run (dart:async/zone.dart:1670:54)
#2      _FutureListener.handleWhenComplete (dart:async/future_impl.dart:247:18)
#3      Future._propagateToListeners.handleWhenCompleteCallback (dart:async/future_impl.dart:872:39)
#4      Future._propagateToListeners (dart:async/future_impl.dart:928:11)
#5      Future._addListener.<anonymous closure> (dart:async/future_impl.dart:513:9)
#6      _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
#7      _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)
#8      _runPendingImmediateCallback (dart:isolate-patch/isolate_patch.dart:118:13)
#9      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:185:5)
```

### DIFF
--- a/pkgs/async/lib/src/result/error.dart
+++ b/pkgs/async/lib/src/result/error.dart
@@ -53,7 +53,7 @@ class ErrorResult implements Result<Never> {
       errorHandler(error);
     } else {
       throw ArgumentError(
-        'is nor Function(Object, StackTrace) neither Function(Object)',
+        'is neither Function(Object, StackTrace) nor Function(Object)',
         'errorHandler',
       );
     }
--- a/pkgs/async/lib/src/result/error.dart
+++ b/pkgs/async/lib/src/result/error.dart
@@ -47,10 +47,15 @@ class ErrorResult implements Result<Never> {
   /// has to be a function expecting only one argument, which will be called
   /// with only the error.
   void handle(Function errorHandler) {
-    if (errorHandler is ZoneBinaryCallback) {
+    if (errorHandler is dynamic Function(Object, StackTrace)) {
       errorHandler(error, stackTrace);
+    } else if (errorHandler is dynamic Function(Object)) {
+      errorHandler(error);
     } else {
-      (errorHandler as ZoneUnaryCallback)(error);
+      throw ArgumentError(
+        'is nor Function(Object, StackTrace) neither Function(Object)',
+        'errorHandler',
+      );
     }
   }
 
--- a/pkgs/async/lib/src/subscription_stream.dart
+++ b/pkgs/async/lib/src/subscription_stream.dart
@@ -76,12 +76,20 @@ class _CancelOnErrorSubscriptionWrapper<T>
     super.onError((Object error, StackTrace stackTrace) {
       // Wait for the cancel to complete before sending the error event.
       super.cancel().whenComplete(() {
-        if (handleError is ZoneBinaryCallback<void, Object, StackTrace> || handleError is ZoneBinaryCallback) {
-          handleError?.call(error, stackTrace);
-        } else if (handleError is ZoneUnaryCallback<void, Object> || handleError is ZoneUnaryCallback) {
-          handleError?.call(error);
+        if (handleError is ZoneBinaryCallback<void, Object, StackTrace>) {
+          handleError(error, stackTrace);
+        } else if (handleError is ZoneBinaryCallback) {
+          handleError(error, stackTrace);
+        } else if (handleError is ZoneUnaryCallback<void, Object>) {
+          handleError(error);
+        } else if (handleError is ZoneUnaryCallback) {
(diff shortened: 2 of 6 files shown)


kevmoo:
A test here would be GREAT

pkgs/async/test/subscription_stream_test.dart
```diff
@@ -148,8 +148,80 @@ void main() {
       });
     }
   });
+
+  group('subscriptionStream error callback', () {
+    test('- binary typed', () async {
```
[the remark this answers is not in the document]
PROGrand: Descriptions changed

kevmoo:
@lrhn – thoughts on this?

pkgs/async/lib/src/subscription_stream.dart
```diff
@@ -76,12 +76,16 @@ class _CancelOnErrorSubscriptionWrapper<T>
...
-          (handleError as ZoneUnaryCallback)(error);
+        } else if (handleError is dynamic Function(Object)) {
+          handleError(error);
         }
       });
     });
   }
 }
+
+abstract class BinaryWrapper {
```
kevmoo, line 89: Are you using this at all?

PROGrand: Sorry, removed unused.

lrhn (approved):

pkgs/async/lib/src/subscription_stream.dart
```diff
@@ -76,12 +76,16 @@ class _CancelOnErrorSubscriptionWrapper<T>
     super.onError((Object error, StackTrace stackTrace) {
       // Wait for the cancel to complete before sending the error event.
       super.cancel().whenComplete(() {
-        if (handleError is ZoneBinaryCallback) {
+        if (handleError is dynamic Function(Object, StackTrace)) {
```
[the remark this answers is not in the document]
PROGrand: Yes,
Initially i fixed this as:
```dart
        if (handleError is ZoneBinaryCallback<void, Object, StackTrace>) {
          handleError(error, stackTrace);
        } else if (handleError is ZoneBinaryCallback) {
          handleError(error, stackTrace);
        } else if (handleError is ZoneUnaryCallback<void, Object>) {
          handleError(error);
        } else if (handleError is ZoneUnaryCallback) {
          handleError(error);
        }
```
but `handleError is dynamic Function(Object, StackTrace)` from platform seems to be more compact and unversal

PROGrand:
@lrhn 
Should `Error::handle` to be changed also?


https://github.com/dart-lang/core/blob/3b35aca9a651b63e7405e578800c4a6cfa91bbae/pkgs/async/lib/src/result/error.dart#L49-L55

lrhn:
That code does look just as wrong. Good catch.
It hasn't been correct since before Dart 2.0 (it actually worked as intended in Dart 1.0).
It's a little scary that nobody has noticed until now. Either nobody types their error handlers (because they rely on inference, which they don't get) or nobody uses the code at all.

PROGrand:
> That code does look just as wrong. Good catch. It hasn't been correct since before Dart 2.0 (it actually worked as intended in Dart 1.0). It's a little scary that nobody has noticed until now. Either nobody types their error handlers (because they rely on inference, which they don't get) or nobody uses the code at all.

Ok, i will change this piece too.

pkgs/async/lib/src/result/error.dart
```diff
@@ -47,10 +47,15 @@ class ErrorResult implements Result<Never> {
...
   void handle(Function errorHandler) {
-    if (errorHandler is ZoneBinaryCallback) {
+    if (errorHandler is dynamic Function(Object, StackTrace)) {
       errorHandler(error, stackTrace);
+    } else if (errorHandler is dynamic Function(Object)) {
+      errorHandler(error);
     } else {
-      (errorHandler as ZoneUnaryCallback)(error);
+      throw ArgumentError(
+        'is nor Function(Object, StackTrace) neither Function(Object)',
```
kevmoo, line 56:
```suggestion
        'is neither Function(Object, StackTrace) nor Function(Object)',
```

kevmoo (approved):

PR #859, opened 2025-02-12
Collection: Major version, use 3.0 language features.

lrhn:
Move `package:collection` to using Dart 3.0 language features, in particular `final` and `interface` classes.

Remove deprecated APIs.
Deprecate a few new APIs for various reasons (like better versions being available in the platform libraries).
Includes a potentially breaking change to `SetEquality` and `MapEquality` which should increase performance.

Adding `final` or `interface` to classes requires a major version increment.
That will be *incredibly painful* to land since `package:test` depends on collection `^1.x.0` and everything in the world, this package included, depends on `package:test`. (Half of that world depends on `package:collection` directly as well, it's not just `test`.)

For now (to be able to run tests), the pubspec version is `1.x.0-2.0.0.wip`.

(The `CanonicalizedMap` class is only marked `@sealed`, not `final`, until an existing subclass can be fixed.
It's unknown how many other packages will be broken by these class modifiers.)


pkgs/collection/lib/src/iterable_zip.dart
```diff
@@ -13,7 +13,9 @@ import 'dart:collection';
 /// combined into a single list, which becomes the next value of this
 /// [Iterable]'s [Iterator]. As soon as any of the iterators run out,
 /// the zipped iterator also stops.
-class IterableZip<T> extends IterableBase<List<T>> {
+/// @nodoc
+@Deprecated('Use [i1, i2].zip from dart:collection')
+final class IterableZip<T> extends IterableBase<List<T>> {
```
kevmoo, line 18: should we just do another "dot" release to deprecate and then drop this? Let's double down on the breaking changes.

lrhn:
Do not expect this to be able to land as-is.
The good question is "if ever". Can we, in any possible way, create a 2.0 of `package:collection` when the entire ecosystem uses `^1.x.0` constraints?

Jonas suggested not actually removing anything, just keeping it deprecated *and* `@nodoc`.
Adding `final` is still breaking. We can use `@sealed` instead. Don't have an annotation for `interface`.
It means we'll likely never get to using `final`.

And the changes to `SetEquality` might still be breaking anyway.

jakemac53:
Please mark this PR as a Draft if you don't intend to land it :).

I would definitely argue we shouldn't do this, it is better to just keep the tech debt of the deprecated things forever.

Or, we could delete them in a non-breaking change. I also think that would be better, even though it violates semver.

jakemac53:
(or an even better alternative, convince flutter to stop pinning deps, or at least have a process for not pinning certain specific deps for specified time periods)

lrhn:
Any major version increment of a widely used package is hard. 

It probably requires a lot of packages to change their dependencies to `>=1.19.0 <3.0.0` when they have validated that they work with both versions.
Then, when everyone has done that, individual packages can start requiring the 2.0.0 version without breaking anything. 

That's what we simulate when we fiddle with version constraints of the SDK when publishing a new major version.

Semver doesn't help. It can't, a major version increment is a breaking change, you have to opt in to it.

The alternative is really just to release `collection2` as an unrelated package.

Or allow Pub to include two different versions of the same package in one program, so one package can use collection 1.19.0, and another 2.0.0.
That requires package resolution to be configurable per-package.
Could choose to only allow one version pet major version. 
(But that's not a short-term fix.)

We could create `collection2` as a new package, and `collection` v2.0.0 exporting that (hard version constraint). Then if we ever manage to get people moved to collection 2.0, we can swap the packages so that `collection2` exports `collection` v2, and deprecate `collection2`.

lrhn:
I would find it incredibly worrisome if we choose to make a breaking change as a minor version increment because we can't figure out how to release a major version increment. 
That would be saying that semver doesn't work, and rather than fixing how you use semver (like allowing versions of the same package in one program), we just ignore it and lie about whether the change is breaking.
That may just incentivize people to accept smaller version ranges, rather than `^1.X.0` (since that would no longer mean "Until next breaking change").

Maybe we need a new concept: deprecation-aware semver.
If a package version 2.0.0 can say that it is _deprecation compatible_ with versions 1.17.0 and above, then pub could choose to solve with 2.0.0 of all dependencies require 1.17 or above, and can therefore be assumed to not have deprecation warnings for deprecations that were in those versions.
Like a compatibility window, if you don't use anything deprecated in the version you require, then you will be compatible with any later version that is deprecation-compatible with that version. 

We'll have to have more kinds of deprecations if we want to make every later API change, like "this parameter will stop being optional" or "... will change type to", or "this class will become final/interface/base/sealed".

Probably can't get away with "this enum will get more values".

jakemac53:
> It probably requires a lot of packages to change their dependencies to `>=1.19.0 <3.0.0` when they have validated that they work with both versions.

The problem is that since flutter pins package:collection, the majority of users (and possibly packages, but I am not sure what percentage of these packaged depend on flutter) can't get a version solve for version 2 until flutter uses it. But we don't want flutter to use it until most packages support it, so its a catch 22.

They can use a dependency override on package:collection, but do we want to tell >3k package authors to add a dependency override on a package for this purpose, and pre-emptively publish? We would need to communicate widely here what we expect package authors to do, and coordinate across the whole ecosystem to make this happen.

And then the question becomes, what is the benefit we get in return for all that effort across not only our team but the entire package ecosystem, and is that a good investment.

> I would find it incredibly worrisome if we choose to make a breaking change as a minor version increment because we can't figure out how to release a major version increment.
> That would be saying that semver doesn't work, and rather than fixing how you use semver (like allowing versions of the same package in one program), we just ignore it and lie about whether the change is breaking.

I understand where you are coming from here, and usually I would agree. But, we need to take a step back and think about _why_ we use semver, what the tradeoffs are, and ultimately what will best serve our users. Semver is just a tool - a tool that we use in order to avoid the pain of package version incompatibilities and breakage. However, it also comes with its own pain, which is that when you do a breaking change all consumers of your package have to take some action, even if they are not affected by the change.

I would argue that when a package is core enough, and the breaking change is actually unlikely to anybody in practice, it is sometimes the best thing for our users to violate semver.

I don't like it, but its important we try to evaluate which option is actually the least painful. And a breaking change in a core package like this is incredibly painful and costly.

devoncarew:
> > Or allow Pub to include two different versions of the same package in one program, so one package can use collection 1.19.0, and another 2.0.0.
> >
> > That requires package resolution to be configurable per-package.
> > Could choose to only allow one version pet major version.
> 
> I very much think we should do this!

fwiw, I believe this is how Rust solves the diamond dependency problem. It allows the solve, but provisions two versions of the package locally. Things would get tricky if you ever start passing types from those packages around your app - TypeA from package.v1 would not be the same type as TypeA from package.v2.

nex3:
If you minimize the scope of breaking changes to things that are unlikely to be problems in practice and make all Dart Team packages use `>=1.x.0 <3.0.0` constraints, that should create a window of opportunity where Flutter can relatively painlessly change their pinned version to 2.x.

kekavc24:
If this is still alive, is it possible include some missing methods to `UnmodifiableMapMixin`. It's missing `addEntries`, `removeWhere`, `update`, `updateAll` methods. (I'm sorry I know this additional work on top of the brainstorming about the release strategy and way forward)

sigurdm:
> Isn't the point of Unmodifiable to not have those methods? Or am I being dense?

Well, they have to be mixed in to something that ends up implementing `List` no? So they might as well provide a throwing implementation.

mosuem:
> And to repeat myself - why do we even have this package? Why not merge any useful stuff into dart:collection and discontinue this package?

At least philosophically, I think we should try to move as much as possible _out_ of the SDK. The more is in packages, the better.

> I (as already said) don't see that we gain much from this.

For this PR specifically, I don't have any opinion if it's worth it. But generally we shouldn't hold back with bumping major versions. It won't break the ecosystem, just take longer to propagate.

jonasfj:
Avoid breaking changes in popular packages is not the same as letting them
go stale.

That said, breaking changes today incur a lot of cost. We do have way to
migrate using constraints spanning two major versions, but the process is
slow and for popular packages it requires a lot actors to (a) take action
and update their package, and, (b) be careful with the constraints and what
functionality they depend on.

That's not impossible, but the cost is high. Thus, IMO the benefits from a
breaking change should be commensurate.

If we're just removing deprecated classes, maybe it's better to hide them
from dartdoc. If we're just adding final to class, maybe @sealed is a
better compromise.

If we had better tooling for gradual migrations, such as
private_dependencies, the cost would be lower, and the benefit required for
a breaking change would not have to be as high.

I don't have fully considered thoughts on this, and it's certainly hard to
balance whether or not the benefit of breaking outweighs the cost.

ons. 8. okt. 2025, 09.29 skrev Moritz ***@***.***>:

PR #866, opened 2025-02-27
Adjust the implements clause of IntX

eernstg:
Adjust the implements clause of IntX.

See https://github.com/dart-lang/sdk/issues/59776 for breaking change processing.

Currently, `IntX` implements `Comparable<Object>`. This causes the equation `X extends Comparable<X>` to have no solution which is a supertype of `IntX` (and, hence, `Int64` and `Int32`). This is inconvenient because it causes invocations of methods like the extension method `sortedBy` to fail: There is no actual type argument that satisfies the constraints.

With this PR, `IntX` is a subtype of `Comparable<IntX>`, which means that said methods can now be invoked. Type inference fails when the language version is below 3.7.0 (so the type argument `IntX` must be provided explicitly), but from 3.7.0 and up the feature 'inference-using-bounds' is enabled by default, and type inference will then succeed.

### DIFF
--- a/pkgs/fixnum/CHANGELOG.md
+++ b/pkgs/fixnum/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.2.0
+
+* Change `IntX` such that it implements `Comparable<IntX>`. This makes it
+  possible for `IntX` (and in turn `Int64` and `Int32`) be used with methods
+  like `sortedBy` in the platform libraries.
+
 ## 1.1.1
 
 * Require Dart `^3.1.0`
--- a/pkgs/fixnum/CHANGELOG.md
+++ b/pkgs/fixnum/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.2.0
+## 1.2.0-wip
 
 * Change `IntX` such that it implements `Comparable<IntX>`. This makes it
   possible for `IntX` (and in turn `Int64` and `Int32`) be used with methods
--- a/pkgs/fixnum/pubspec.yaml
+++ b/pkgs/fixnum/pubspec.yaml
@@ -1,5 +1,5 @@
 name: fixnum
-version: 1.1.1
+version: 1.2.0
 description: >-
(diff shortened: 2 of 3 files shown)


eernstg:
The test coverage check fails. The page about test coverage states that "Test coverage should not decrease by introducing a PR", but I can't see how this PR would have that effect. @devoncarew, can you say something about how is this usually handled?

devoncarew:
> The test coverage check fails. The page about test coverage states that "Test coverage should not decrease by introducing a PR", but I can't see how this PR would have that effect. @devoncarew, can you say something about how is this usually handled?

The coverage check is informational. It looks like the health check is failing, and likely due to a bug in the api check tool: https://github.com/dart-lang/core/actions/runs/13566672156/job/37921655020?pr=866#step:12:159.

That check is trying to diff this current change vs. the last published one, and determine what type of semver bump should be done (major? minor? service?). I'll add a label to skip the health check for this PR.

pkgs/fixnum/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.2.0
```
devoncarew, line 1: I think a minor version bump for this change is reasonable? But let's change this to a WIP; we'll roll the change through the sdk and google3 before we publish, i order to flush out any issues.

```suggestion
## 1.2.0-wip
```

pkgs/fixnum/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: fixnum
-version: 1.1.1
+version: 1.2.0
```
devoncarew, line 2:
```suggestion
version: 1.2.0-wip
```

devoncarew (commented):
Will leave the approve to @lrhn -

eernstg:
Thanks!

@lrhn, do you have further comments?

lrhn (approved):

PR #875, opened 2025-04-02
[collection] explicitly make BoolList abstract interface

kevmoo:
It couldn't be extended in practice since it doesn't have a
public constructor

Also made the only implementations final

### DIFF
--- a/pkgs/collection/lib/src/boollist.dart
+++ b/pkgs/collection/lib/src/boollist.dart
@@ -10,7 +10,7 @@ import 'unmodifiable_wrappers.dart' show NonGrowableListMixin;
 /// A space-efficient list of boolean values.
 ///
 /// Uses list of integers as internal storage to reduce memory usage.
-abstract /*mixin*/ class BoolList with ListMixin<bool> {
+abstract interface class BoolList with ListMixin<bool> {
   static const int _entryShift = 5;
 
   static const int _bitsPerEntry = 32;
@@ -180,7 +180,7 @@ abstract /*mixin*/ class BoolList with ListMixin<bool> {
   }
 }
 
-class _GrowableBoolList extends BoolList {
+final class _GrowableBoolList extends BoolList {
   static const int _growthFactor = 2;
 
   _GrowableBoolList._withCapacity(int length, int capacity)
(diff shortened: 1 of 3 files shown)


pkgs/collection/lib/src/boollist.dart
```diff
@@ -10,7 +10,7 @@ import 'unmodifiable_wrappers.dart' show NonGrowableListMixin;
 /// A space-efficient list of boolean values.
 ///
 /// Uses list of integers as internal storage to reduce memory usage.
-abstract /*mixin*/ class BoolList with ListMixin<bool> {
+abstract interface class BoolList with ListMixin<bool> {
```
lrhn, line 13: I'd make it `final`. (Could be `sealed`, but nobody should care that there are exactly two implementations.)

There is zero reason to ever `implements BoolList`, it has the same interface as `List<bool>`.
The fact that you *can* shouldn't be taken as encouragement.

(We don't have `@Deprecate.interface()` yet, so there is no good way to deprecate implementability.)

lrhn (approved):
LGTM, but go for `final` if possible.

pkgs/collection/lib/src/boollist.dart
kevmoo: Went with `@sealed` for now, with a note about future breaking change.

pkgs/collection/lib/src/boollist.dart
```diff
@@ -175,12 +178,16 @@ abstract /*mixin*/ class BoolList with ListMixin<bool> {
     }
   }
 
+  /// Note: [index] is NOT checked for validity.
+  bool _getBit(int index) =>
```
kevmoo, line 182: FYI: seems far too complex to not centralize. Concerns about not being inlined @lrhn ?

lrhn: Not sure what "centralize" means here - if anything this is centralized in that all bit-reads use this one implementation. 
I hope it's inlined, but I leave that to compilers. They (probably) know better.

For performance, it's likely a little inefficient because computed shifts are expensive on x64 chips.
But everybody does it this way, so hopefully it'll be fine.

pkgs/collection/lib/src/boollist.dart
```diff
@@ -167,6 +169,7 @@ abstract /*mixin*/ class BoolList with ListMixin<bool> {
   @override
   Iterator<bool> get iterator => _BoolListIterator(this);
 
+  /// Note: [index] is NOT checked for validity.
```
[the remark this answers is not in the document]
kevmoo: Were you going for prescriptive there or descriptive?

PR #880, opened 2025-04-16
add support to ignore ansi sequences when formatting usage display. Fixes #879

### LINKED ISSUE #879: ANSI styling sequences within the `valueHelp` parameter of addOption() causes incorrect usage display
If you include ANSI styling sequences in `valueHelp` parameters of addOption() is causes the usage display to display incorrect formatting because it is including the hidden ANSI escape sequences in the string length calculations.

Example

Image

The correct display would be:

Image

timmaffett:
This PR adds logic to exclude hidden ANSI escape sequences when formatting the Usage display.
The ANSI escape sequences are excluded when calculating the lengths of help strings used in the Usage display by using a string extension `lengthWithoutAnsi` which removes any hidden ANSI escape sequences present before calculating the string length.

It includes tests of the `lengthWithoutAnsi` getter to ensure that it returns correct length values when no ANSI escape sequences are present as well as when a variety of ANSI escape sequences are present.

(The regex that i am using here is well exercised, as it is the same reg ex that I use within VS Code when parsing and formatting ANSI sequences within the debug console)

---

- [X ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/args/lib/src/utils.dart
+++ b/pkgs/args/lib/src/utils.dart
@@ -3,9 +3,38 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'dart:math' as math;
 
+
+/// A utility class for finding and stripping ANSI codes from strings.
+class _AnsiUtils {
+  static final String ansiCodePattern = [
+      '[\\u001B\\u009B][[\\]()#;?]*(?:(?:(?:[a-zA-Z\\d]*(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]*)*)?\\u0007)',
+      '(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PR-TZcf-ntqry=><~]))'
+    ].join('|');
+
+  static final RegExp ansiRegex = RegExp(ansiCodePattern);
+
+  static String stripAnsi(String source) {
+    return source.replaceAll(ansiRegex, '');
+  }
+
+  static bool hasAnsi(String source) {
+    return ansiRegex.hasMatch(source);
+  }
+}
+
+/// A utility extension on [String] to provide ANSI code stripping and length
+/// calculation without ANSI codes.
+extension StringUtils on String {
+  /// Returns the length of the string without ANSI codes.
+  int get lengthWithoutAnsi {
+    if (!_AnsiUtils.hasAnsi(this)) return length;
+    return _AnsiUtils.stripAnsi(this).length;
+  }
+}
+
 /// Pads [source] to [length] by adding spaces at the end.
 String padRight(String source, int length) =>
-    source + ' ' * (length - source.length);
+    source + ' ' * (length - source.lengthWithoutAnsi);
 
 /// Wraps a block of text into lines no longer than [length].
 ///
--- a/pkgs/args/test/utils_test.dart
+++ b/pkgs/args/test/utils_test.dart
@@ -16,6 +16,12 @@ final _indentedLongLineWithNewlines =
 const _shortLine = 'Short line.';
 const _indentedLongLine = '    This is an indented long line that needs to be '
     'wrapped and indentation preserved.';
+const _ansiReset = 'This is normal text. \x1B[0m<- Reset point.';
+const _ansiBoldTextSpecificReset = 'This is normal, \x1B[1mthis is bold\x1B[22m, and this uses specific reset.';
+const _ansiMixedStyles = 'Normal, \x1B[31mRed\x1B[0m, \x1B[1mBold\x1B[0m, \x1B[4mUnderline\x1B[0m, \x1B[1;34mBold Blue\x1B[0m, Normal again.';
+const _ansiLongSequence = 'Start \x1B[1;3;4;5;7;9;31;42;38;5;196;48;5;226m Beaucoup formatting! \x1B[0m End';
+const _ansiCombined256 = '\x1B[1;38;5;27;48;5;220mBold Bright Blue FG (27) on Gold BG (220)\x1B[0m';
+const _ansiCombinedTrueColor = '\x1B[4;48;2;50;50;50;38;2;150;250;150mUnderlined Light Green FG on Dark Grey BG\x1B[0m';
 
 void main() {
   group('padding', () {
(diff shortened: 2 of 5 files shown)


munificent:
Sorry, but I definitely don't have time right now to maintain args. :(

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,38 @@
...
+}
+
+/// A utility extension on [String] to provide ANSI code stripping and length
+/// calculation without ANSI codes.
+extension StringUtils on String {
+  /// Returns the length of the string without ANSI codes.
+  int get lengthWithoutAnsi {
+    if (!_AnsiUtils.hasAnsi(this)) return length;
+    return _AnsiUtils.stripAnsi(this).length;
+  }
```
lrhn, line 32: This is more expensive than it needs to be.
You know the length of the total string, and you have the matches for every Ansi escape sequence, so you could just subtract the lengths of that, without needing to actually create a new string.
```dart
  int get lengthWithoutAnsi {
    var length = this.length;
    for (var escape in _AnsiUtils.ansiRE.allMatches(this)) {
      length -= escape.end - escape.start;
    }
    return length;
  }
```
(If we are a little lucky, the RegExp match won't eagerly allocate a match string if we don't ask for it.)

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,38 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'dart:math' as math;
 
+
+/// A utility class for finding and stripping ANSI codes from strings.
+class _AnsiUtils {
+  static final String ansiCodePattern = [
+      '[\\u001B\\u009B][[\\]()#;?]*(?:(?:(?:[a-zA-Z\\d]*(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]*)*)?\\u0007)',
+      '(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PR-TZcf-ntqry=><~]))'
+    ].join('|');
```
lrhn, line 12: Use raw strings for RegExp patterns. No need to use lists and `join`, just have adjacent strings:
```dart
  static const String ansiCodePattern = 
     r'[\x1b\x9b][[\]()#;?]*'
     r'(?:'  //
     r'(?:'  // -
     r'(?:[a-zA-Z\d]*' //--
     r'(?:;[\-a-zA-Z\d/#&.:=?%@~_]*)*'
     r')?\x07' // -
     r'|'
     r'(?:' // --
     r'(?:\d{1,4}(?:;\d{0,4})*[\dA-PR-TZcf-ntqry=><~]'
     r')' //-
     r')'; //
```

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,38 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'dart:math' as math;
 
+
+/// A utility class for finding and stripping ANSI codes from strings.
+class _AnsiUtils {
+  static final String ansiCodePattern = [
```
lrhn, line 9: It would be very nice to have a textual description of what the pattern is supposed to match.
It's hard enough to debug RegExps even if you have a specification of what they're intended to do.
Without that, it's impossible to tell whether it's right or wrong.

Is this a correct description:
* One of U+001b or U+009b
* Followed by any number of the characters: `[`, `]`, `(`, `)`, `#`, `;`, `?`
* Followed by either:
    * zero or more letters or digits
    * followed by any number of repeats of:
        * `;`
        * followed by any number of digits, letters, or characters from `-/#^.:=?%@~_`
* or:
    *  1-4 digits
    * followed by any number of repeats of:
        *  `;`
        * followed by 0-4 digits
    * followed by on digit, letter other than QUVWXY or abdeopsuvwxz, or one of `=><~`.

I don't know what the official specification of this is.
I'd be tempted to just go with what Wikpedia says is a simple control sequence introduction:

`r'\x1b[0-?]*[ -/]*[@-~]'`

Using U+009B as well is dangerous. It means something else in both UTF-8 and Windows-1252 code pages.

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,38 @@
...
+  int get lengthWithoutAnsi {
+    if (!_AnsiUtils.hasAnsi(this)) return length;
+    return _AnsiUtils.stripAnsi(this).length;
+  }
+}
+
 /// Pads [source] to [length] by adding spaces at the end.
 String padRight(String source, int length) =>
-    source + ' ' * (length - source.length);
+    source + ' ' * (length - source.lengthWithoutAnsi);
```
lrhn, line 37: Use `padRight`: `source.padRight(length - source.lengthWithoutAnsi)`.

That should avoid doing an extra allocation of just the spaces .

pkgs/args/test/utils_test.dart
```diff
@@ -16,6 +16,12 @@ final _indentedLongLineWithNewlines =
 const _shortLine = 'Short line.';
 const _indentedLongLine = '    This is an indented long line that needs to be '
     'wrapped and indentation preserved.';
+const _ansiReset = 'This is normal text. \x1B[0m<- Reset point.';
+const _ansiBoldTextSpecificReset = 'This is normal, \x1B[1mthis is bold\x1B[22m, and this uses specific reset.';
+const _ansiMixedStyles = 'Normal, \x1B[31mRed\x1B[0m, \x1B[1mBold\x1B[0m, \x1B[4mUnderline\x1B[0m, \x1B[1;34mBold Blue\x1B[0m, Normal again.';
+const _ansiLongSequence = 'Start \x1B[1;3;4;5;7;9;31;42;38;5;196;48;5;226m Beaucoup formatting! \x1B[0m End';
+const _ansiCombined256 = '\x1B[1;38;5;27;48;5;220mBold Bright Blue FG (27) on Gold BG (220)\x1B[0m';
+const _ansiCombinedTrueColor = '\x1B[4;48;2;50;50;50;38;2;150;250;150mUnderlined Light Green FG on Dark Grey BG\x1B[0m';
```
lrhn, line 24: These are all of the form `\x1b\[[\d;]*m`. That doesn't cover most of the complexity of the RegExp.

If this is all the functionality needs to do, the RegExp shouldn't be as complicated. (It probably shouldn't, even if recognizing more escapes than just the color ones.)

pkgs/args/lib/src/utils.dart
timmaffett: I now use a much simpler regular expression that just matches all possible ANSI escape codes, WITHOUT validating the escape codes.   I have also provided an explanation of the much simpler regex.

```
/// Matches the Control Sequence Introducer (CSI) ANSI escape sequences.
///
/// Anatomy:
/// \x1b     : The literal ESC character (ASCII 27).
/// \[       : The literal '[' character (together with ESC, this forms the CSI).
/// [0-9;?]* : Zero or more parameter bytes:
///             - 0-9 : Numeric parameters (e.g., color codes).
///             - ;   : Parameter separators.
///             - ?   : Private mode indicators (e.g., cursor toggles).
/// [a-zA-Z] : The 'Final Byte' that determines the command (e.g., 'm' for color).
static final RegExp _ansiRegex = RegExp(r'\x1b\[[0-9;?]*[a-zA-Z]');
```

pkgs/args/lib/src/utils.dart
timmaffett: The simpler regex now uses raw string

timmaffett:
> Hey @timmaffett could you address the review feedback if you are interested in moving forward on this?

Hi @brianquinlan / @lrhn - I am sorry it took me so long to get back to your feedback - I have now implemented all of the suggested changes and it greatly simplifies things.  The tests have also been extended to test the regex and new .ansiLength extension thoroughly.

Do I need to open another PR or can use RE-open this one ?

Issue #872, opened 2025-03-19
package:args shouldn't throw ArgumentError on missing argument


eseidel:
ArgumentError is an Error and thus supposed to be used for programmer errors.

ArgResults.operator[] throws ArgumentError on what should be a runtime error:
https://github.com/dart-lang/core/blob/61e677100b06d56a1b3731ab1178ebf9102ecb1f/pkgs/args/lib/src/arg_results.dart#L69

Am I wrong?

RandalSchwartz:
I could argue that Error subclasses are for things that the programmer should have prevented through proper coding but failed to do so.  But this error seems to be correct code and a problem with the data.  I could easily argue that this should be an API Exception rather than a programmer Error.

lrhn:
You're wrong 😉 
At least on that line. Now, line 74 on the other hand ... 

In line 69 it checks if the name the programmer passes in is one that the same programmer has defined in the parser options for the parser. If there is no defined option with that name, then it's not a name of an option at all.

This particular argument parsing is based on defining all the valid flags and options up front, then parsing a command line based on that, and then looking up the result. If the lookup uses a flag name that isn't defined, that's a programmer error.

In line 74, an `ArgumentError` is thrown for a data issue
https://github.com/dart-lang/core/blob/61e677100b06d56a1b3731ab1178ebf9102ecb1f/pkgs/args/lib/src/arg_results.dart#L73C5-L75C6

The option was defined, as required even, but the parsed result contains no data for it.
There are two possibilities here:
- This can't happen, parsing would have thrown instead (which would "the way"!), but then it shouldn't be throwing an `ArgumentError`, just have an `assert` or throw an `AssertionError`.
- It can happen, and then it should throw an `Exception` or have a different return value.

eseidel:
Confirmed:

Line 74 is hit:
https://github.com/dart-lang/core/blob/61e677100b06d56a1b3731ab1178ebf9102ecb1f/pkgs/args/lib/src/arg_results.dart#L74

```
import 'package:args/args.dart';
import 'package:test/test.dart';

void main() {
  test('test', () {
    final argParser =
        ArgParser()
          ..addFlag('help', help: 'Show this help message.')
          ..addFlag('verbose', help: 'Show verbose output.')
          ..addOption('bar', help: 'The bar option.')
          ..addOption('baz', help: 'The baz option.', mandatory: true);

    final arguments = <String>[];

    final results = argParser.parse(arguments);

[...]

PR #888, opened 2025-05-05
fix: ArgResults.operator[] should throw Exception not Error

eseidel:
ArgResults.operator[] and ArgResults.option() would throw ArgumentError instead of Exception when the user would fail to pass all expected arguments.

Errors are only supposed to be thrown on programmer error so changed to throw ArgParserException instead.

Fixes https://github.com/dart-lang/core/issues/872

This is a breaking change!


devoncarew:
> I definitely don't have time to actively maintain args.

FYI, you're getting these reviews due to a recent change to the https://github.com/dart-lang/core/blob/main/.github/CODEOWNERS file; if you'd prefer not to get them, feel free to update that file (likely to change the reviewers to the ecosystem team).

> I'm honestly not sure if this is worth landing, as there is a non-trivial risk of breaking lots of apps in the wild.

I suspect that the mandatory feature probably shouldn't have landed; this use of errors instead of exceptions was introduced then. We have a `next-breaking-release` label; it could serve to queue this change up for consideration later.

Separately, I wonder if the least overall disruption for users is changing the exception type, or, removing the `mandatory` option? Removing the option would let people know that they had work to update their code.

munificent:
> FYI, you're getting these reviews due to a recent change to the https://github.com/dart-lang/core/blob/main/.github/CODEOWNERS file; if you'd prefer not to get them, feel free to update that file (likely to change the reviewers to the ecosystem team).

Oh, well how about that. Sent a PR: https://github.com/dart-lang/core/pull/891

> I suspect that the mandatory feature probably shouldn't have landed; this use of errors instead of exceptions was introduced then. We have a `next-breaking-release` label; it could serve to queue this change up for consideration later.

Yeah, I didn't love it when I first saw that go by. Definitely a well-intentioned feature, but it doesn't fit very well with existing API for how args works.

> Separately, I wonder if the least overall disruption for users is changing the exception type, or, removing the `mandatory` option? Removing the option would let people know that they had work to update their code.

The least painful option is to probably leave things as they are with the current exception type. If we do want to support mandatory options, we should probably re-work the API so that you can't get your hands on an `ArgResults` at all if any of the mandatory options are missing. In other words, validate that all mandatory options are present during parsing time, and not when options are accessed from results.

brianquinlan:
I've added the "next-breaking-release" label so that we can consider this if/when we get to args 3.0

PR #890, opened 2025-05-07
[path] Fix test race condition caused by changing Directory.current

### LINKED ISSUE #889: Running "dart test" inside "core/pkgs/path" has race condition and sporadic failures
Running `dart test` inside "core/pkgs/path" on my machine fails several time per 10 runs.

Obviously the concurrency matters - for me `-j` defaults to `6`.

Applying this change makes the thing happen more often (for me, like this it happens every time):

```
diff --git a/pkgs/path/test/io_test.dart b/pkgs/path/test/io_test.dart
index e428b89a..c51ca7ab 100644
--- a/pkgs/path/test/io_test.dart
+++ b/pkgs/path/test/io_test.dart
@@ -97,8 +97,12 @@ void main() {
     () {
       final dir = path.current;
       try {
+        io.sleep(const Duration(seconds: 2));
+
         io.Directory.current = path.rootPrefix(path.current);
 
+        io.sleep(const Duration(seconds: 2));
+
         expect(
           path.relative(path.absolute('foo/bar'), from: path.current),
           path.relative(path.absolute('foo/bar')),
diff --git a/pkgs/path/test/path_map_test.dart b/pkgs/path/test/path_map_test.dart
index 11c4a3e0..8b9f84b7 100644
--- a/pkgs/path/test/path_map_test.dart
+++ b/pkgs/path/test/path_map_test.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
+
 import 'package:path/path.dart';
 import 'package:test/test.dart';
 
@@ -70,10 +72,19 @@ void main() {
     });
 
     test('uses the second value in the case of duplicates', () {
-      final map = PathMap.of({'foo': 1, absolute('foo'): 2});
+      final orgPath = Directory.current.path;
+      final absolute2 = absolute('fooSpecial');
+      if (Directory.current.path != orgPath) {
+        // ignore: only_throw_errors
+        throw 'Who knows which it used.';
+      }
+      while(Directory.current.path == orgPath) {
+        sleep(const Duration(milliseconds: 1));
+      }
+      final map = PathMap.of({'fooSpecial': 1, absolute2: 2});
       expect(map, hasLength(1));
-      expect(map, containsPair('foo', 2));
-      expect(map, containsPair(absolute('foo'), 2));
+      expect(map, containsPair('fooSpecial', 2));
+      expect(map, containsPair(absolute('fooSpecial'), 2));
     });
   });
 }
```

(I run `dart test -j2` - but it probably just has to be `j>1`).

jensjoha:
Fixes #889.

### DIFF
--- a/pkgs/path/test/io_test.dart
+++ b/pkgs/path/test/io_test.dart
@@ -43,81 +43,78 @@ void main() {
 
     test(
       'uses the previous working directory if deleted',
-      () {
-        final dir = io.Directory.current.path;
-        try {
-          final temp = io.Directory.systemTemp.createTempSync('path_test');
-          final tempPath = temp.resolveSymbolicLinksSync();
-          io.Directory.current = temp;
+      currentDirHelper(() {
+        final temp = io.Directory.systemTemp.createTempSync('path_test');
+        final tempPath = temp.resolveSymbolicLinksSync();
+        io.Directory.current = temp;
 
-          // Call "current" once so that it can be cached.
-          expect(path.normalize(path.absolute(path.current)), equals(tempPath));
+        // Call "current" once so that it can be cached.
+        expect(path.normalize(path.absolute(path.current)), equals(tempPath));
 
-          temp.deleteSync();
+        temp.deleteSync();
 
-          // Even though the directory no longer exists, no exception is thrown.
-          expect(path.normalize(path.absolute(path.current)), equals(tempPath));
-        } finally {
-          io.Directory.current = dir;
-        }
-      },
+        // Even though the directory no longer exists, no exception is thrown.
+        expect(path.normalize(path.absolute(path.current)), equals(tempPath));
+      }),
       //TODO: Figure out why this is failing on windows and fix!
       skip: io.Platform.isWindows ? 'Untriaged failure on Windows' : false,
     );
   });
 
-  test('registers changes to the working directory', () {
+  test('registers changes to the working directory', currentDirHelper(() {
     final dir = io.Directory.current.path;
-    try {
-      expect(path.absolute('foo/bar'), equals(path.join(dir, 'foo/bar')));
-      expect(
-        path.absolute('foo/bar'),
-        equals(path.context.join(dir, 'foo/bar')),
-      );
+    expect(path.absolute('foo/bar'), equals(path.join(dir, 'foo/bar')));
+    expect(
+      path.absolute('foo/bar'),
+      equals(path.context.join(dir, 'foo/bar')),
+    );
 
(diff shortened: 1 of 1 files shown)


jensjoha:
When reviewing do use the "hide whitespace" option as most of the "change" is whitespace changes caused by the removal of the try/finally.

pkgs/path/test/io_test.dart
```diff
@@ -43,81 +43,78 @@ void main() {
...
-        io.Directory.current = dir;
-      }
-    },
+    }),
     //TODO(kevmoo): figure out why this is failing on windows and fix!
     skip: io.Platform.isWindows ? 'Untriaged failure on Windows' : null,
   );
 }
+
+dynamic Function() currentDirHelper(dynamic Function() body) {
```
lrhn, line 113: Can do `R Function() currentDirHelper<R>(R Function() body) {` to avoid the `dynamic`. (But if it's always used in a context with no static type constraint, it doesn't matter. Could use `Object?` instead of `dynamic` then.)

Or it could just return `void`, and then you use it inside the test, instead of around the test callback:
```dart
test('....', () {
  currentDirHelper(() {
    // Actual body.
  });
}
```

lrhn, line 113: Consider naming it `withLocalCurrentDirectory` or `withLocalCWD`.
Document it to say:
```dart
/// Runs [body] in a zone with a local current working directory.
///
/// Avoids clobbering the current working directory of the entire process
/// when writing to it and reading it back through `dart:io` functions.
```
Or something.

lrhn (approved):

pkgs/path/test/io_test.dart
jensjoha: Done.

pkgs/path/test/io_test.dart
```diff
@@ -43,81 +43,78 @@ void main() {
...
+    );
+  }));
+
+  // Regression test for #35. This tests against the *actual* working directory
+  // rather than just a custom context because we do some processing in
+  // [path.current] that has clobbered the root in the past.
+  test(
+    'absolute works on root working directory',
+    currentDirHelper(() {
+      io.sleep(const Duration(seconds: 2));
```
jensjoha: Ooops. I used this while testing, and left it in. Removed.

pkgs/path/test/io_test.dart
jensjoha: Done.

PR #894, opened 2025-05-26
Optimize surrogate decoding.

lrhn:
Use `char ^ 0xD800 <= 0x3FF` to check if a char code is a lead surrogate. That avoids doing a later `& 0x3FF` to get rid of the top bits. Similar for tail surrogate.

This ensures that the `high` function gets values without high bits, which makes it smaller (it tries to get inlined, so a little smaller counts).

Also optimize that function to reduce dependency depth and try to hit `base + (something < small)` expressions that can optimized into a single x64 address computation.

Gives a ~7% increase on backwards traversal and 30% increase for forward traversal, based on tool/benchmark.dart compiled with `dart compile exe`.
Actually a small decrease in performance on web for forward iteration, and a small increase for backwards iteration, and Wasm follows Web in performance here.

(Also found a bug in the generator, which hasn't worked since it was last committed.)

Interestingly, the change makes little-to-no difference on the `benchmark/benchmark.dart` benchmark.
(Maybe even makes it a little slower.)

### DIFF
--- a/pkgs/characters/tool/benchmark.dart
+++ b/pkgs/characters/tool/benchmark.dart
@@ -18,8 +18,7 @@ void main(List<String> args) {
   if (args.isNotEmpty) {
     count = int.parse(args[0]);
   }
-  var text =
-      genesis +
+  var text = genesis +
       hangul +
       genesis +
       diacretics +


pkgs/characters/lib/src/grapheme_clusters/table.dart
```diff
@@ -1139,10 +1139,10 @@ int low(int codeUnit) {
 @pragma('vm:prefer-inline')
 @pragma('wasm:prefer-inline')
 int high(int lead, int tail) {
-  var offset = (((0x3ff & lead) << 10) + (0x3ff & tail)) + (2048 << 8);
-  var chunkStart = _start.codeUnitAt(offset >> 8);
-  var index = chunkStart + (tail & 255);
-  return _data.codeUnitAt(index);
+  var offset = (tail >> 8) + (lead << 2);
```
mosuem, line 1142: IIUC, this assumes that `tail` and `lead` don't need to be masked with `0x3ff`. Should this be `assert`ed here?

mosuem (approved):
LGTM - Although I don't have much knowledge of this package, and don't understand its intricacies. But what I understand makes sense to me.

PR #897, opened 2025-06-13
Add support for emitting usage as a JSON schema

natebosch:
- Vendor a copy of a library defining a JSON schema API which can output
  as a `Map<String, Object?>`.
- Add a `jsonSchema` getter on `ArgParser` which is similar to `usage`
  but outputs a schema instead of help text.

Clients can manually support a `--json-help` or similar argument in the
same way they support `--help`.

### DIFF
--- a/pkgs/args/lib/src/arg_parser.dart
+++ b/pkgs/args/lib/src/arg_parser.dart
@@ -6,6 +6,7 @@ import 'dart:collection';
 
 import 'allow_anything_parser.dart';
 import 'arg_results.dart';
+import 'json_schema.dart';
 import 'option.dart';
 import 'parser.dart';
 import 'usage.dart';
@@ -392,4 +393,48 @@ class ArgParser {
   /// Finds the option whose name or alias matches [name], or `null` if no
   /// option has that name or alias.
   Option? findByNameOrAlias(String name) => options[_aliases[name] ?? name];
+
+  Map<String, Object?> get jsonSchema {
+    final properties = <String, Schema>{};
+    final required = <String>[];
+    for (final option in _options.values) {
+      if (option.hide) continue;
+      var help = option.help;
+      final extras = <String>[];
+      if (option.defaultsTo != null) {
+        extras.add('defaults to "${option.defaultsTo}"');
+      }
+      if (option.allowed?.isNotEmpty ?? false) {
+        extras.add('allowed values: ${option.allowed?.join(', ')}');
+      }
+      if (extras.isNotEmpty) {
+        help = [
+          if (help != null) help,
+          ...extras,
+        ].join('\n');
+      }
+      final schema = switch (option.type) {
+        OptionType.flag => Schema.bool(
+            description: help,
+          ),
+        OptionType.single => Schema.string(
+            description: help,
+          ),
+        OptionType.multiple => Schema.list(
+            description: help,
+            items: Schema.string(),
+          ),
+        _ => throw StateError('Unhandled Option Type: ${option.type.name}')
+      };
+
+      if (option.mandatory) {
+        required.add(option.name);
+      }
+      properties[option.name] = schema;
+    }
+    return Schema.object(
+      properties: properties,
+      required: required,
+    ).asMap();
+  }
 }
--- a/pkgs/args/lib/src/json_schema.dart
+++ b/pkgs/args/lib/src/json_schema.dart
@@ -1,0 +1,611 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Copied from https://github.com/dart-lang/ai/blob/failures-only/pkgs/dart_mcp/lib/src/api/tools.dart
+
+/// The valid types for properties in a JSON-RCP2 schema.
+enum JsonType {
+  object('object'),
+  list('array'),
+  string('string'),
+  num('number'),
+  int('integer'),
+  bool('boolean'),
+  nil('null');
+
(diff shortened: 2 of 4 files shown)


pkgs/args/lib/src/json_schema.dart
```diff
@@ -0,0 +1,611 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Copied from https://github.com/dart-lang/ai/blob/failures-only/pkgs/dart_mcp/lib/src/api/tools.dart
```
jakemac53, line 5:
```suggestion
// Copied from https://github.com/dart-lang/ai/blob/main/pkgs/dart_mcp/lib/src/api/tools.dart
```

jakemac53 (approved):

pkgs/args/lib/src/json_schema.dart
```diff
@@ -0,0 +1,611 @@
...
+/// {
+///   "productId": "ABC12345",
+///   "productName": "Super Widget",
+///   "price": 19.99,
+///   "tags": ["electronics", "gadget"],
+///   "dimensions": {"length": 10, "width": 5, "height": 2.5}
+/// }
+/// ```
+///
+/// And invalid data (e.g., missing productName, or an extra undefined
```
lrhn, line 192: `productName` -> `` `"productName"` ``

pkgs/args/lib/src/json_schema.dart
```diff
@@ -0,0 +1,611 @@
...
+    Map<String, Schema>? patternProperties,
+    List<String>? required,
+
+    /// Must be one of bool, Schema, or Null
+    Object? additionalProperties,
+    bool? unevaluatedProperties,
+    StringSchema? propertyNames,
+    int? minProperties,
+    int? maxProperties,
+  }) =>
```
lrhn, line 223: (Could also be redirecting constructor `: this.fromMap(...);`. No big difference.)

pkgs/args/lib/src/json_schema.dart
```diff
@@ -0,0 +1,611 @@
...
+      ObjectSchema.fromMap({
+        'type': JsonType.object.typeName,
+        if (title != null) 'title': title,
+        if (description != null) 'description': description,
+        if (properties != null) 'properties': properties,
+        if (patternProperties != null) 'patternProperties': patternProperties,
+        if (required != null) 'required': required,
```
lrhn, line 230: With null-aware elements:
```dart
ObjectSchema.fromMap({
        'type': JsonType.object.typeName,
        'title': ?title,
        'description': ?description,
        'properties': ?properties,
        'patternProperties': ?patternProperties,
        'required': ?required,
        'additionalProperties': ?additionalProperties,
        'unevaluatedProperties': ?unevaluatedProperties,
        'propertyNames': ?propertyNames,
        'minProperties': ?minProperties,
        'maxProperties': ?maxProperties,
      });
```

pkgs/args/lib/src/arg_parser.dart
```diff
@@ -392,4 +393,48 @@ class ArgParser {
...
+      if (option.mandatory) {
+        required.add(option.name);
+      }
+      properties[option.name] = schema;
+    }
+    return Schema.object(
+      properties: properties,
+      required: required,
+    ).asMap();
+  }
```
lrhn, line 439: So basically just:
```dart
  var properties = <String, Object?>{};
  var required = [];
  for (var option in _options.values) {
    if (option.hide) continue;
    var Option(:name, :type, :help) = option;
    final description = [?help,
      if (option.defaultsTo != null)  'defaults to "${option.defaultsTo}"'
      if (option.allowed?.isNotEmpty ?? false) 
         'allowed values: ${option.allowed?.join(', ')}'
    ].join('\n');
    properties[name] = switch (type) {
      OptionType.bool => {
          'type': 'boolean',
          'description': description,
      },
      OptionType.single => {
         'type': 'string',
          'description': description,
       },
       OptionType.multi => {
         'type': 'list',
         'items': const {'type': 'string'}
         'description': description,
       },
    };   
    if (option.mandatory) required.add(name);
  }  
  return {
    'type': 'object',
    'properties': properties,
    if (required.isNotEmpty) 'required': required
  };
}  
```  

I'm not sure a complete JSONSchema implementation is worth it, especially one we didn't write ourselves, but is vendoring and now have to maintain anyway.

natebosch:
> Has there been a request for this?

I used this to create a schema for the test runner which we can use in our MCP server.
https://github.com/dart-lang/ai/pull/169

I don't know if it makes sense to land this here since it might be breaking, but I at least wanted a reference implementation on a shared branch.

> If we don't accept options being _input_ as JSON, why are we giving a JSON schema for them?

I had suggested adding support for a JSON argument file to a few folks in a meeting and no one thought it was a good idea. I don't have strong feelings about it either way, it's pretty easy to convert. We could move the utility which does the conversion here if we want that part shared.

```dart
extension on Map<String, Object?> {
  Iterable<String> asCliArgs() sync* {
    for (final MapEntry(:key, :value) in entries) {
      if (value is List) {
        for (final element in value) {
          yield '--$key';
          yield element as String;
        }
        continue;
      }
      yield '--$key';
      if (value is bool) continue;
      yield value as String;
    }
  }
}
```

https://github.com/dart-lang/ai/pull/169/files#diff-c3420b6d2d3f69afb4dbb2c88882dc2a2d324dad52cf5853b3fd23b6c9436529R269

natebosch:
I think the trajectory of LLM tooling has moved away from this type of schema being useful. I'll close for now and we can revisit if demand resurfaces.

PR #903, opened 2025-08-31
feat(core): Add intersperse method to Iterable

abbashosseinii:
This PR introduces the intersperse extension method to Iterable, which inserts a separator element between each element of the original sequence. This provides an efficient, lazy, and declarative way to achieve this common pattern.

For example, [1, 2, 3].intersperse(0) would produce an iterable yielding 1, 0, 2, 0, 3.

This addresses the feature request in issue #49900.

- [x] I've reviewed the contributor guide and applied the relevant portions to this PR.

<details>
<summary>Contribution guidelines</summary>

See our [contributor guide](https://www.google.com/search?q=https://github.com/dart-lang/sdk/blob/main/CONTRIBUTING.md) for general expectations for PRs.

Larger or significant changes should be discussed in an issue before creating a PR.

Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use dart format.

Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://www.google.com/search?q=https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md%23making-a-change).

Changes to packages require [corresponding tests](https://www.google.com/search?q=https://github.com/dart-lang/sdk/blob/main/CONTRIBUTING.md%23testing).

</details>

### DIFF
--- a/pkgs/collection/CHANGELOG.md
+++ b/pkgs/collection/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 1.20.0-wip
 
+- Add `intersperse` method to `IterableExtension` for efficient element separation.
 - Add `IterableMapEntryExtension` for working on `Map` as a list of pairs, using
   `Map.entries`.
 - Explicitly mark `BoolList` as `abstract interface`


pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -613,6 +613,33 @@ extension IterableExtension<T> on Iterable<T> {
...
+  ///
+  /// Example:
+  /// ```dart
+  /// final numbers = [1, 2, 3];
+  /// final withSeparators = numbers.intersperse(0);
+  /// print(withSeparators.toList()); // Output: [1, 0, 2, 0, 3]
+  /// ```
+  ///
+  /// The resulting iterable is lazy. It only computes the next value when
+  /// it's requested.
```
lrhn: I wouldn't say this. That's what iterables are expected to do.
A function returning an `Iterable` should document if it *doesn't* work that way.

pkgs/collection/test/extensions_test.dart
```diff
@@ -1224,6 +1224,33 @@ void main() {
...
+        test('single element', () {
+          expect(iterable([1]).intersperse(0), [1]);
+        });
+        test('two elements', () {
+          expect(iterable([1, 2]).intersperse(0), [1, 0, 2]);
+        });
+        test('multiple elements', () {
+          expect(iterable([1, 2, 3, 4]).intersperse(0), [1, 0, 2, 0, 3, 0, 4]);
+        });
+        test('with different separator types', () {
```
lrhn: Different from what?
(Not from the element type, which is the interesting case.)

pkgs/collection/test/extensions_test.dart
```diff
@@ -1224,6 +1224,33 @@ void main() {
...
+          expect(iterable(['a', 'b', 'c']).intersperse('-'),
+              ['a', '-', 'b', '-', 'c']);
+          expect(iterable([1.5, 2.5, 3.5]).intersperse(0.0),
+              [1.5, 0.0, 2.5, 0.0, 3.5]);
+        });
+        test('with nullable int separator', () {
```
lrhn: Is not nullable?

PR #909, opened 2025-09-10
Fix short flag parsing to strip equals sign from value

### LINKED ISSUE #908: [args] Short flag with equals syntax incorrectly includes equals sign in parsed value
There's an inconsistency in how the args package parses short flags versus long flags when using equals syntax. Short flags include the equals sign in the parsed value, while long flags correctly exclude it.

Current Behavior
 
- `--option=value` correctly parses as `"value"`
- `-o=value` parses as `"=value"`

Reproduction
```dart
  import 'package:args/args.dart';

  void main() {
    final parser = ArgParser();
    parser.addOption('option', abbr: 'o');

    final result1 = parser.parse(['--option=value']);
    final result2 = parser.parse(['-o=value']);

    print('Long flag:  --option=value "${result1['option']}"');  // Output: "value"
    print('Short flag: -o=value "${result2['option']}"');  // Output: "=value"

    // These should be equal but they're not
    assert(result1['option'] == result2['option']);  // Fails
  }
```

Arpi223:
When parsing short flags with equals syntax (e.g., -o=value), the parser was incorrectly including the equals sign in the parsed value, resulting in '=value' instead of 'value'. This was inconsistent with long flag parsing (--option=value) which correctly returned 'value'.

This fix strips the leading equals sign when present, making short and long flag parsing consistent.

This fixes issues in tools like FlutterFire CLI where users expect -i=bundleid to work the same as --ios-bundle-id=bundleid.

Fixes #908 
---

<details>

</details>

### DIFF
--- a/pkgs/args/lib/src/parser.dart
+++ b/pkgs/args/lib/src/parser.dart
@@ -207,6 +207,11 @@ class Parser {
       // The first character is a non-flag option, so the rest must be the
       // value.
       var value = '${lettersAndDigits.substring(1)}$rest';
+      // If the value starts with '=' (like in -o=value), strip it to match
+      // the behavior of long options (--option=value).
+      if (value.startsWith('=')) {
+        value = value.substring(1);
+      }
       _setOption(_results, first, value, '-$c');
     } else {
       // If we got some non-flag characters, then it must be a value, but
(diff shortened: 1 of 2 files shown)


davidmorgan:
As commented on the issue--I think this is working as intended.

PR #911, opened 2025-09-22
Fix `Int64.operator ==`

### LINKED ISSUE #910: Int64.== shouldn't throw an exception
It tries to `_promote` the value and fails, when an object has an unrelated type:

```
Invalid argument (other): not an int, Int32 or Int64: Instance of 'NullValue'
package:fixnum/src/int64_native.dart 185:5                                               Int64._promote
package:fixnum/src/int64_native.dart 273:43                                              Int64.==
```

osa1:
Note: `operator ==` implementations in this package are not symmetric, but we consider that as intentional. See #910 for details.

Fixes #910.

### DIFF
--- a/pkgs/fixnum/test/int32_test.dart
+++ b/pkgs/fixnum/test/int32_test.dart
@@ -214,18 +214,25 @@ void main() {
     });
 
     test('==', () {
-      expect(Int32(17), isNot(equals(Int32(18))));
-      expect(Int32(17), equals(Int32(17)));
-      expect(Int32(17), isNot(equals(Int32(16))));
-      expect(Int32(17), isNot(equals(Int64(18))));
-      expect(Int32(17), equals(Int64(17)));
-      expect(Int32(17), isNot(equals(Int64(16))));
-      expect(Int32.MIN_VALUE, isNot(equals(Int32.MAX_VALUE)));
-      expect(Int32(17), isNot(equals(18)));
-      expect(Int32(17) == 17, isTrue); // ignore: unrelated_type_equality_checks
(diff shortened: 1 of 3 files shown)


osa1:
I don't think this requires a changelog entry as it fixes a bug in an unreleased change that's already in the changelog.

pkgs/fixnum/test/int32_test.dart
```diff
@@ -214,18 +214,25 @@ void main() {
...
-      expect(Int32(17), isNot(equals(Int64(18))));
-      expect(Int32(17), equals(Int64(17)));
-      expect(Int32(17), isNot(equals(Int64(16))));
-      expect(Int32.MIN_VALUE, isNot(equals(Int32.MAX_VALUE)));
-      expect(Int32(17), isNot(equals(18)));
-      expect(Int32(17) == 17, isTrue); // ignore: unrelated_type_equality_checks
-      expect(Int32(17), isNot(equals(16)));
-      expect(Int32(17), isNot(equals(Object())));
-      expect(Int32(17), isNot(equals(null)));
+      // Note: do not use `equals` matcher below as it considers exceptions as
```
lrhn, line 217: Should still be able to use it for equals, just not for `isNot(equals(...))`.

But I guess it's easier to just use the same approach everywhere.

lrhn: Even then, exceptions in `==` should still cause a mismatch in `equals`, so the test still fails and should throw a test failure, instead of the exception that `==` threw itself. (But `isNot(equals...)` won't throw, and that's a problem.)

Issue #628, opened 2021-12-21
Add alternateWith extension method on Iterable


alexeyinkin:
There is a repeating feature request in Flutter to add separators for children of Row and Column:
https://github.com/flutter/flutter/issues/16957
https://github.com/flutter/flutter/issues/18917
https://github.com/flutter/flutter/issues/55378
https://github.com/flutter/flutter/issues/95263
https://github.com/flutter/flutter/issues/95521

At its highest it got 117 thumbs-ups.
But it is declined over and over again because it can be solved outside of the framework with a few lines of code:

```dart
Iterable<T> alternateWith(Iterable<T> items, T separator) {
  return items
      .expand((item) sync* { yield separator; yield item; })
      .skip(1);
}
```

Can we add it to this package as an extension on Iterable? One could then do:

```dart
Column(
  children: <Widget>[
    Widget1(),
    Widget2(),
  ].alternateWith(separator).toList(),
),
```

One could easily make another package with this, as there are many. Yet they are unknown and may be poorly maintained. This package is widely used, endorsed, and can be referred to in Flutter docs thus ending the recurrance of the feature request or speeding up the debate should it occur again.

I can do it myself if approved.

lrhn:
The design space is large enough that I don't want to lock anyone into any choice.

For myself, I'd probably do a single function that can insert before, after and between, defaulting to between:
```dart
extension IterableSeparators<T> on Iterable<T>
  Iterable<T> separate(T separator, {bool before = false, bool after = false, bool between = true})  sync* {
    if (before) yield separator;
    var first = true;
    for (var element in this) {
      if (between && !first) yield separator;
      yield element;
      first = false;
    }
    if (after && !(first && before)) yield separator;
  }
}
```

But then I'd also have a different version if I want to build a list directly, instead of creating an intermediate 
iterable. 
Even a version that inserts separators into an existing list, like this one:
```dart
extension ListSeparators<T> on List<T> {
  void insertSeparators(T separator, {bool before = false, bool after = false, bool between = true}) {
    var length = this.length;
    if (length == 0) {
      if (before || after) add(separator);
      return;
    }
    if (!between) {
      if (before) insert(0, separator);
      if (after) add(separator);
      return;
   }
    var newLength = length * 2 - 1;
    if (before) newLength++;
    if (after) newLength++;
    var offset = before ? 1 : 0; 
    // New position of element at index i is i*2+offset.
    // Element at new position `index` is:
[...]

PR #919, opened 2025-10-31
Add `separated`, `separatedList` and `separate` to iterables and lists.

lrhn:
Each new function separates the elements of an iterable or list with a separator value of the same type.
They allow optionally adding the separator before a first element and/or after a last element.

* `separated`: Returns a (lazy) iterable backed by the original elements.
* `separatedList`: Creates an (eager) list with the same elements that separated would have returned. (But more efficiently.) (Also has a version specialized for a `List` input.)
* `separate`: Modifies a list in-place to add separators around the existing elements.

Fixes, among others, https://github.com/dart-lang/core/issues/628

This is one possible API. Names are fungible. Options are optional.

Attempts to give you the most direct API for what you want, so you don't have to create an intermediate iterable just to call `toList()` on it, and not call `toList()` if you could just modify the original list. 

WDYT?

### DIFF
--- a/pkgs/collection/lib/src/iterable_extensions.dart
+++ b/pkgs/collection/lib/src/iterable_extensions.dart
@@ -56,6 +56,102 @@ extension IterableExtension<T> on Iterable<T> {
     return chosen;
   }
 
+  /// The elements of this iterable separated by [separator].
+  ///
+  /// Emits the same elements as this iterable, and also emits
+  /// a [separator] between any two of those elements.
+  ///
+  /// If [before] is set to `true`, a [separator] is also
+  /// emitted before the first element.
+  /// If [after] is set to `true`, a [separator] is also
+  /// emitted after the last element.
+  ///
+  /// If this iterable is empty, [before] and [after] have no effect.
+  ///
+  /// Example:
+  /// ```dart
+  /// print([1, 2, 3].separated(-1)); // (1, -1, 2, -1, 3)
+  /// print([1].separated(-1)); // (1)
+  /// print([].separated(-1)); // ()
+  ///
+  /// print([1, 2, 3].separated(
+  ///   -1
+  ///   before: true,
+  /// )); // (-1, 1, -1, 2, -1, 3)
+  ///
+  /// print([1].separated(
+  ///   -1
+  ///   before: true,
+  ///   after: true,
+  /// )); // (-1, 1, -1)
+  ///
+  /// print([].separated(
+  ///   -1
+  ///   before: true,
+  ///   after: true,
+  /// )); // ()
+  /// ```
+  Iterable<T> separated(T separator,
+      {bool before = false, bool after = false}) sync* {
+    const emitBefore = 1;
+    const emitAfter = 2;
+    var state = before ? emitBefore : 0;
+    final afterState = emitBefore | (after ? emitAfter : 0);
+    for (var element in this) {
+      if (state & emitBefore != 0) yield separator;
+      state = afterState;
+      yield element;
+    }
+    if (state & emitAfter != 0) yield separator;
+  }
+
+  /// Creates new list with the elements of this list separated by [separator].
+  ///
+  /// Returns a new list which contains the same elements as this list,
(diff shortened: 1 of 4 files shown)


lrhn:
That does look rather convincing. I had hoped that we could optimize a list `for-in` more, but if we don't inline and the loop is polymorphic, then that won't happen. And the body is more complicated to account for the difference between the first and following items.

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -56,6 +56,102 @@ extension IterableExtension<T> on Iterable<T> {
...
+  /// )); // (-1, 1, -1)
+  ///
+  /// print([].separated(
+  ///   -1
+  ///   before: true,
+  ///   after: true,
+  /// )); // ()
+  /// ```
+  Iterable<T> separated(T separator,
+      {bool before = false, bool after = false}) sync* {
```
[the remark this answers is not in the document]
lrhn: (I did write it. Let me try putting it in an see how it feels. The iterator itself is simple. The complications come from me trying to optimize `Iterable` methods too.)

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -56,6 +57,99 @@ extension IterableExtension<T> on Iterable<T> {
...
+  ///
+  /// Returns a new list which contains the same elements as this list,
+  /// with a [separator] between any two of those elements.
+  ///
+  /// If [before] is set to `true`, a [separator] is also
+  /// added before the first element.
+  /// If [after] is set to `true`, a [separator] is also
+  /// added after the last element.
+  ///
+  /// If this list is empty, [before] and [after] have no effect.
```
natebosch:
```suggestion
  /// Creates new list with the elements of this iterable separated by [separator].
  ///
  /// Returns a new list which contains the same elements as this iterable,
  /// with a [separator] between any two of those elements.
  ///
  /// If [before] is set to `true`, a [separator] is also
  /// added before the first element.
  /// If [after] is set to `true`, a [separator] is also
  /// added after the last element.
  ///
  /// If this iterable is empty, [before] and [after] have no effect.
```

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -56,6 +57,99 @@ extension IterableExtension<T> on Iterable<T> {
...
+  ///
+  /// Emits the same elements as this iterable, and also emits
+  /// a [separator] between any two of those elements.
+  ///
+  /// If [before] is set to `true`, a [separator] is also
+  /// emitted before the first element.
+  /// If [after] is set to `true`, a [separator] is also
+  /// emitted after the last element.
+  ///
+  /// If this iterable is empty, [before] and [after] have no effect.
```
natebosch: Do you have any particular use cases in mind for the `before` and `after` arguments where not marking the boundaries of an empty iterable are the expected choice?

I don't necessarily doubt that this set of arguments/behaviors is the most useful one to provide, but I can't convince myself that it's the most useful behavior and it looks like it makes the Iterable implementation a fair bit more complex.

pkgs/collection/lib/src/iterable_extensions.dart
```diff
@@ -1056,3 +1150,206 @@ extension ComparatorExtension<T> on Comparator<T> {
...
+  @override
+  Iterable<T> take(int count) {
+    if (count == 0) return Iterable<T>.empty();
+    var beforeCount = _flags >> 1;
+    if (count == 1) {
+      if (beforeCount == 0) {
+        return _elements.take(1);
+      }
+      // return Iterable<T>.value(_separator); // Why you no exist?!
+      return DelegatingIterable<T>([_separator]);
```
natebosch: Are we wrapping this because we're worried about a caller detecting it's a list and starting to rely on that detail?

natebosch (approved):

lucavenir:
when will this land on stable? 👀

is the naming definitive?

thanks in advance 🙏🏼

cedvdb:
still unreleased

PR #923, opened 2025-11-05
Add use_null_aware_elements to recommended

### LINKED ISSUE #899: Add use_null_aware_elements to recommended lint set
I was surprised to find that when I upgraded Dart, I wasn't getting any lints to use the new null aware elements feature, then I realized I had to turn on the lint myself even after upgrading the lints package. I think this would be good to add to at least the recommended lint set.

> and lints that enforce writing Dart using a single, idiomatic style and format.

I think this lint aligns with the goal of the recommended set, and encourages more people to use the new feature.

There is an open false negative issue about this lint so that should be fixed before this is ever landed: dart-lang/sdk#61122

natebosch:
Closes #899

Bump to a feature release.
Drop an unnecessary `-0` suffix on the SDK constraint.
Update URL for reading rules from the dart site source.

### DIFF
--- a/pkgs/lints/tool/gen_docs.dart
+++ b/pkgs/lints/tool/gen_docs.dart
@@ -27,7 +27,7 @@ const rulesMarkdownFilePath = 'rules.md';
 
 /// Fetches the [rulesUrl] JSON description of all lints, saves a cached
 /// summary of the relevant fields in [rulesCacheFilePath], and
-/// updates [rulesMarkdownFilePath] to
+/// updates [rulesMarkdownFilePath].
 ///
 /// Passing any command line argument disables generating documentation,
 /// and makes this tool just verify that the doc is up-to-date with the


pkgs/lints/tool/gen_docs.dart
```diff
@@ -10,8 +10,10 @@ import 'package:yaml/yaml.dart' as yaml;
 import 'package:http/http.dart' as http;
 
 /// Source of truth for linter rules.
-const rulesUrl =
-    'https://raw.githubusercontent.com/dart-lang/site-www/main/src/_data/linter_rules.json';
+final rulesUri = Uri.https(
+  'raw.githubusercontent.com',
+  'dart-lang/site-www/main/src/data/linter_rules.json',
+);
```
leafpetersen: Unrelated nit, feel free to ignore, but the comment on line 30 below ends prematurely (or has a hanging "to" that should be deleted).

PR #936, opened 2026-02-05
Platform 3.2

lrhn:
The `Platform.current` object exposes one (or potentially more)
of a `NativePlatform` (based on `dart:io`) or `BrowserPlatform`
(based on JS `window.navigator` or anything else relevant).

Each has its own API, and the current platform determines
whether a value of these types is available or not.
On unsupported platforms, a value is just `null`.

Provides a `testing.dart` library which exposes
`TestNativePlatform` and `TestBrowserPlatform`,
which are customizable classes that can be used for testing.
It's possible to run code with such a fake value as
the *current* platform (`Platform.current` is temporarily
updated to a value which exposes the, fx, `TestNativePlatform`
as its `Platform.nativePlatform`).

Still retains the legacy and deprecated API that makes
native properties directly available on `Platform`.
Code should migrate to the new API by doing:
* `const LocalPlatform()` &rarr; `Platform.current.nativePlatform!`
* `new FakePlatform()` &rarr; `new TestNativePlatform()`

It's currently not possible to have more than one non-`null`
platform-value.
Platforms are detected by the presence of the `dart:io` and `dart:js_interop`
libraries, and no current platform has both.
When such a platform exists, the code will need to be adapted to to produce
*both* (possible using conditional imports or conditional parts
and augmentations).


pkgs/platform/example/bin/platform_using_switch.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Example of safely accessing browser platform properties.
```
natebosch: [nit]
```suggestion
/// Example of safely accessing platform specific properties.
```

pkgs/platform/README.md
```diff
@@ -1,4 +1,4 @@
-[![Pub](https://img.shields.io/pub/v/platform.svg)](https://pub.dartlang.org/packages/platform)
+[![Pub](https://img.shields.io/pub/v/platform.svg)https://pub.dartlang.org/packages/platform)
```
kevmoo:
```suggestion
[![Pub](https://img.shields.io/pub/v/platform.svg)](https://pub.dartlang.org/packages/platform)
```

parlough:
```suggestion
[![Pub](https://img.shields.io/pub/v/platform.svg)](https://pub.dev/packages/platform)
```

pkgs/platform/lib/src/platform_specific/unknown_platform.dart
```diff
@@ -0,0 +1,106 @@
...
+import 'non_io_native_platform.dart';
+import 'non_web_browser_platform.dart';
+
+export 'non_io_native_platform.dart';
+export 'non_web_browser_platform.dart';
+
+const platformInstance = Platform._();
+
+// For the legacy `LocalPlatform`.
+const NativePlatform? nativePlatformInstance = null;
```
natebosch: [nit] Should this be deprecated? Or is there any way we should mark it to ensure it gets cleaned up after a major version bump?

pkgs/platform/tool/check_migration.dart
```diff
@@ -0,0 +1,117 @@
...
+    if (verbose > 0) {
+      stdout.writeln('Running: dart analyze ${testFile.path}');
+    }
+    var dartAnalyzeResult = run(dartExe, ['analyze', testFile.path]);
+    stdout.writeln(dartAnalyzeResult.stdout);
+    // TODO: Maybe, if verbose, and `/bin/diff` exists, print the diff after
+    // fixing.
+    // var diffres = Process.runSync("/bin/diff", [sourceFilePath, testFile.path],
+    //     runInShell: true);
+    // print(diffres.stdout);
```
natebosch: [nit] delete commented out code if we aren't going to land it for now.

natebosch (approved):

pkgs/platform/tool/check_tree_shaking.dart
```diff
@@ -0,0 +1,259 @@
...
+        ['compile', 'js', '-o', jsOutput, example.path],
+      );
+      if (!_checkCompiled(example, File(jsOutput), verbose)) {
+        printProcessOutput(result);
+        continue;
+      }
+      if (!check(jsOutput, exampleJS)) {
+        failures.add(exampleJS);
+        // Try finding `d8` in path, and d8.js preamble in SDK.
+        // Until then, make sure to have a `d8` in then path which adds
```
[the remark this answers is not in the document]
lrhn: I think it actually works without the preamble, so just having `d8` in path is enough.
Will add.

pkgs/platform/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+  platform information, or get a `null` on non-native platforms.
+  Similarly `Platform.current.web` is only non-`null` on the web.
+
+  Mocking APIs are moved to a separate library to ensure they are
+  only used for testing, and to allow better performance in production.
+
+  Legacy APIs are deprecated, but retained and emulated using the new API.
+  Use `dart fix` to migrate off the deprecated API.
+
 ## 3.1.7-wip
```
kevmoo, line 15: Was 3.1.7 ever published?

PR #938, opened 2026-02-09
Add `ArgResult.flagCount`.

### LINKED ISSUE #937: Allow flags to occur multiple times, and reveal how many.
It's not possible to determine whether a flag was passed more than once. (Or I can't find it.)

A flag is either present or absent, and `ArgResult.flag` returns a `bool`.
The `callback` is only called once, even if the flag is present multiple times.

There is no way to distinguish `-vv` from `-v`, where the former is a known and used format for increasing a verbosity level by two.

What is needed is a flag-based equivalent of a "multi-option".

There is no need to introduce a new `addMultiFlag`, any flag can count its positive occurrences, and `ArgResult.flag(name)` can just return whether than number is greater than zero. 
Then there can be an `ArgResult.flagCount(name)` that returns this number.

A `--no-the-flag` resets the flag count to zero.

lrhn:
Counts the occurrences of a flag, and makes it accessible in the `ArgResult`.

Only changes parser and result class, the `Option` class still treats flags as `bool` options, with a value that is `true` only when the count is greater than zero.

Fixes #937.

(The `-vv` is something I often use in my own scripts, and one reason for never using the `args` package.)

### DIFF
--- a/pkgs/args/lib/src/arg_results.dart
+++ b/pkgs/args/lib/src/arg_results.dart
@@ -66,15 +66,18 @@ class ArgResults {
   /// > flags, [option] for options, and [multiOption] for multi-options.
   dynamic operator [](String name) {
     if (!_parser.options.containsKey(name)) {
-      throw ArgumentError('Could not find an option named "--$name".');
+      throw ArgumentError.value(
+          name, 'name', 'Could not find an option named "--$name".');
     }
 
     final option = _parser.options[name]!;
     if (option.mandatory && !_parsed.containsKey(name)) {
-      throw ArgumentError('Option $name is mandatory.');
+      throw ArgumentError.value(name, 'name', 'Option $name is mandatory.');
     }
-
-    return option.valueOrDefault(_parsed[name]);
+    var parsedValue = _parsed[name];
+    if (option.isFlag && parsedValue is int) parsedValue = parsedValue > 0;
+    var result = option.valueOrDefault(parsedValue);
+    return result;
   }
 
   /// Returns the parsed or default command-line flag named [name].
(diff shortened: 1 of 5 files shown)


pkgs/args/lib/src/arg_results.dart
```diff
@@ -83,12 +86,45 @@ class ArgResults {
...
+    if (parsedValue is int) {
+      parsedValue = parsedValue > 0;
+    }
+    return option.valueOrDefault(parsedValue) as bool;
+  }
+
+  /// The number of times the flag named [name] occurred.
+  ///
+  /// If a flag occurred more than once in the arguments,
+  /// this is the total number of counts.
```
natebosch: [nit]


```suggestion
  /// If a flag occurred more than once in the arguments,
  /// this is the total number of occurrences.
```

pkgs/args/test/parse_test.dart
```diff
@@ -90,6 +90,50 @@ void main() {
...
         throwsIllegalArg(() => results.multiOption('a'));
       });
+
+      test('flagCount', () {
+        var parser = ArgParser();
+        parser.addFlag('flag', abbr: 'f', defaultsTo: false, negatable: false);
+        parser.addFlag('negatable-flag', abbr: 'n', defaultsTo: false);
+
+        // Flags not occurring for testing default value.
+        parser.addFlag('default-true', abbr: 'd', defaultsTo: true);
```
natebosch: [nit] Can you add a test case with a flag that defaults to true and is also passed? Also one that defaults to true and is negated with a `--no` flag.

natebosch (approved):

Levi-Lesches:
Thanks, I was just looking for this when writing a CLI tool

mosuem:
@lrhn this isn't missing much to land IMO. If you find the time to take a short look!

PR #942, opened 2026-02-13
add support to ignore ansi sequences when formatting usage display. Fixes #879

### LINKED ISSUE #879: ANSI styling sequences within the `valueHelp` parameter of addOption() causes incorrect usage display

timmaffett:
This is a RE-push of the #880 PR - the bot closed it and now the code does not update/so I made this PR so that the new changes can be seen.  If the #880 PR got re-opened it would probably update and this could be closed...?

---From ORIGINAL #880 PR ----------------
This PR adds logic to exclude hidden ANSI escape sequences when formatting the Usage display.
The ANSI escape sequences are excluded when calculating the lengths of help strings used in the Usage display by using a string extension lengthWithoutAnsi which removes any hidden ANSI escape sequences present before calculating the string length.

It includes tests of the lengthWithoutAnsi getter to ensure that it returns correct length values when no ANSI escape sequences are present as well as when a variety of ANSI escape sequences are present.

(The regex that i am using here is well exercised, as it is the same reg ex that I use within VS Code when parsing and formatting ANSI sequences within the debug console)


natebosch (changes requested):

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,44 @@
...
+  int get ansiLength {
+    return _ansiRegex
+        .allMatches(this)
+        .fold(0, (sum, match) => sum + match.group(0)!.length);
+  }
+
+  /// Returns the length of the string without ANSI escape sequences.
+  int get lengthWithoutAnsi => length - ansiLength;
+
+  /// Returns the string with all ANSI escape sequences removed.
```
lrhn: (Drive-by-comment, feel free to ignore: _Could_ be a getter named `withoutAnsi`. A function is fine too. It's really a matter of taste and whether the result or the process is considered more important. I'd personally probably go with the getter, I rarely find the process important unless it's very expensive and the reader should be reminded that they're _doing something_. This is an internal utility inside a single package, where the strings are unlikely to be very long. But also ... it's not public API, so I'm nit-picking something nobody will ever care about. This is fine!)

timmaffett: Good point, I agree.

natebosch (approved):

pkgs/args/test/utils_test.dart
```diff
@@ -213,4 +224,311 @@ needs to be wrapped.
...
+          equals(_indentedLongLineWithNewlines.length));
+    });
+    test('lengthWithoutAnsi returns correct length on short line without ansi',
+        () {
+      expect(_shortLine.lengthWithoutAnsi, equals(_shortLine.length));
+    });
+  });
+
+  group('lengthWithoutAnsi is correct with no ANSI sequences', () {
+    test('lengthWithoutAnsi returns correct length on lines without ansi', () {
```
natebosch: Some of these tests appear to be duplicates. Were they copy/pasted?

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,39 @@
...
+  /// * `\x1b`: The literal ESC character (ASCII 27, U+001B).
+  /// * `\[`: The literal `[` character (together with the ESC, this starts the CSI).
+  /// * `[\x30-\x3f]*`: Parameter bytes (`0-9:;<=>?`).
+  /// * `[\x20-\x2f]*`: Intermediate bytes (`!"#$%&'()*+,-./`).
+  /// * `[\x40-\x7e]`: Final byte (`@A-Z[\]^_`a-z{|}~`).
+  static final RegExp _ansiRegExp =
+      RegExp(r'\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]');
+
+  /// Combined length of all ANSI escape sequences in the string.
+  int get ansiLength {
```
natebosch: @lrhn - I'm inclined to make more of these implementation details private and remove the tests of the individual APIs, leaving only the tests for `lengthWithoutAnsi` and `padWithoutAnsi`.  Do you have any concerns?

natebosch (changes requested):
Took a closer look at the tests and I think we may be over-testing.

pkgs/args/test/utils_test.dart
timmaffett: @natebosch 
It was an accidental copy/paste of the "no ANSI sequences" group. I have removed the duplicate.
I also took a broader pass at the over-testing: dropped the tests that exercised the now-private helpers and consolidated the rest, so `utils_test.dart` is down ~115 lines net while keeping the regex edge-case coverage (terminators, private-mode, byte-order, broken/out-of-range sequences) - now done through `lengthWithoutAnsi`.

pkgs/args/lib/src/utils.dart
timmaffett: @natebosch @lrhn  
Done. `ansiLength` is now private (`_ansiLength`), and I removed `withoutAnsi` and `containsAnsi` entirely since nothing outside the tests used them. The public surface is just `lengthWithoutAnsi` and `padRightIgnoreAnsi` (both consumed by `usage.dart`/`command_runner.dart`). The individual-API tests are gone; the edge-case coverage they provided is preserved by re-expressing it in terms of `lengthWithoutAnsi`.

timmaffett:
I believe that I have addressed all the outstanding requested changes.

pkgs/args/lib/src/utils.dart
timmaffett: this has been address - everything private except `lengthWithoutAnsi` and `padRightIgnoreAnsi`

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,44 @@
...
+  }
+
+  /// Returns the length of the string without ANSI escape sequences.
+  int get lengthWithoutAnsi => length - ansiLength;
+
+  /// Returns the string with all ANSI escape sequences removed.
+  String stripAnsi() => replaceAll(_ansiRegex, '');
+
+  /// Returns `true` if the string contains any ANSI escape sequences.
+  bool hasAnsi() {
```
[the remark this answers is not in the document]
timmaffett: no longer exists, general functionality all removed

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,31 @@
...
+  /// * `[\x30-\x3f]*`: Parameter bytes (`0-9:;<=>?`).
+  /// * `[\x20-\x2f]*`: Intermediate bytes (`!"#$%&'()*+,-./`).
+  /// * `[\x40-\x7e]`: Final byte (`@A-Z[\]^_`a-z{|}~`).
+  static final RegExp _ansiRegExp =
+      RegExp(r'\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]');
+
+  /// The combined length of all ANSI escape sequences in the string.
+  int get _ansiLength => _ansiRegExp
+      .allMatches(this)
+      .fold(0, (sum, match) => sum + match[0]!.length);
```
lrhn: Can use `(match.end - match.start)` instead of allocating the string.

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,31 @@
...
+  /// Matches the Control Sequence Introducer (CSI) ANSI escape sequences.
+  ///
+  /// Structure based on ECMA-48:
+  /// * `\x1b`: The literal ESC character (ASCII 27, U+001B).
+  /// * `\[`: The literal `[` character (together with the ESC, this starts the CSI).
+  /// * `[\x30-\x3f]*`: Parameter bytes (`0-9:;<=>?`).
+  /// * `[\x20-\x2f]*`: Intermediate bytes (`!"#$%&'()*+,-./`).
+  /// * `[\x40-\x7e]`: Final byte (`@A-Z[\]^_`a-z{|}~`).
+  static final RegExp _ansiRegExp =
+      RegExp(r'\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]');
```
lrhn: Since this is only used for counting the chars, it can be wrapped in `(?:…)+` to capture adjacent escapes in one match.

pkgs/args/lib/src/utils.dart
```diff
@@ -3,9 +3,31 @@
...
+  int get _ansiLength => _ansiRegExp
+      .allMatches(this)
+      .fold(0, (sum, match) => sum + match[0]!.length);
+
+  /// The length of the string without its ANSI escape sequences.
+  int get lengthWithoutAnsi => length - _ansiLength;
+
+  /// Pads this string to [length] by adding spaces at the end, ignoring
+  /// ANSI escape sequences when calculating the current length.
+  String padRightIgnoreAnsi(int length) => padRight(length + _ansiLength);
```
lrhn: Could take a string to pad with, like `padRight` does, defaulting to `" "`.

lrhn: Could have `padLeft` too, for completeness.

pkgs/args/lib/src/utils.dart
lrhn: Should this be using `lengthWithoutAnsi`?
(And entire function be Ansi-escape aware in general?)

PR #953, opened 2026-04-23
Fix is* OS getters on legacy fake platform

natebosch:
Corrects the behavior of the fields like `isAndroid` when the fake
platform is constructed with `operatingSystem: 'android'`. Existing
tests rely on this behavior.

### DIFF
--- a/pkgs/platform/lib/src/legacy_implementation/legacy_classes.dart
+++ b/pkgs/platform/lib/src/legacy_implementation/legacy_classes.dart
@@ -11,7 +11,8 @@ import 'dart:convert' show JsonDecoder, JsonEncoder;
 
 import '../platforms_impl.dart'
     show BrowserPlatform, NativePlatform, Platform, PlatformTestBase;
-
+// ignore: invalid_use_of_visible_for_testing_member
+import '../testing/test_platforms.dart' show TestNativePlatform;
 // Not showing `lineTerminator` which wasn't used in the legacy code.
 import '../util/json_keys.dart'
     as json_key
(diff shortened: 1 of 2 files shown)


pkgs/platform/lib/src/legacy_implementation/legacy_classes.dart
```diff
@@ -57,6 +58,26 @@ final class FakePlatform extends PlatformTestBase {
   @override
   String? packageConfig;
 
+  @override
+  // ignore: invalid_use_of_visible_for_testing_member
+  NativePlatform get nativePlatform => TestNativePlatform(
```
lrhn, line 63: Instead of creating a new instance on every access, should this just be a final variable initialized by the constructor?

Or really, that should be the only place the values are stored, the `Platform` class shouldn't need it.

Let me try to make a PR that does that.

pkgs/platform/lib/src/legacy_implementation/legacy_classes.dart
```diff
@@ -11,7 +11,8 @@ import 'dart:convert' show JsonDecoder, JsonEncoder;
 
 import '../platforms_impl.dart'
     show BrowserPlatform, NativePlatform, Platform, PlatformTestBase;
-
+// ignore: invalid_use_of_visible_for_testing_member
+import '../testing/test_platforms.dart' show TestNativePlatform;
```
lrhn, line 15: This import is really annoying, since including the `test_platforms` library may make tree-shaking harder.
The legacy classes mix testing and production, so it's probably unavoidable.

The goal is to get rid of the legacy classes in a later (but soon) release, as a major version increment. Then tell people that if they're warning free on 3.2, they can make their dependency be `>= 3.2.0 <5.0.0` and be forward compatible. So this import will go away whent the entire classs does.

pkgs/platform/lib/src/legacy_implementation/legacy_classes.dart
lrhn: And done: https://github.com/dart-lang/core/pull/956

natebosch:
Closing in favor of #956

PR #956, opened 2026-04-24
(Platform) Make `Platform.is<OS>` operations work on legacy `FakePlatform`

lrhn:
Requires importing some testing classes into the main library,
 which is not surprising when the legacy `FakePlatform` class existed for testing to begin with.

### DIFF
--- a/pkgs/platform/lib/platform.dart
+++ b/pkgs/platform/lib/platform.dart
@@ -40,12 +40,13 @@
 /// > and [FakePlatform] classes, and exposes deprecated members on the
 /// > [Platform] interface.
 /// > Code using those deprecated members or the class [LocalPlatform]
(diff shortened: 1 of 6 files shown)


natebosch (approved):

lrhn:
I should merge this into the reapply PR.

natebosch:
> I should merge this into the reapply PR.

I went ahead and pushed this commit to that PR branch.
https://github.com/dart-lang/core/pull/958/changes/226cb993c7ad8f20abdd1bf7f16bd814fa5f139d

I confirmed that this version should be compatible with google3 so we should be good to land it.

natebosch:
Hmm, I'm not sure what was different between my merge when I had this passing in google3 and when I tried it on github, it clearly won't work to copy over the result like I thought it would

PR #959, opened 2026-05-03
Generalize binarySearchBy and lowerBoundBy to accept K key instead of E value

### LINKED ISSUE #895: Binary search and lower bound functions can be generalized
Currently if I have a `List<Foo>` sorted on `String name` field of `Foo`, I can't binary search for or find lower bound of a `String` value, because existing functions require the key to be the same type with the list element type.

For example: https://github.com/dart-lang/core/blob/dc975303c312497aec2f5ed4c5828da3607e33dd/pkgs/collection/lib/src/algorithms.dart#L39-L46

A more general version of this function would be:

```
int binarySearchBy<E, K>(
  List<E> sortedList,
  K Function(E element) keyOf,
  int Function(K, K) compare,
  K key, [    // <------- type of value is K instead of E
  int start = 0,
  int? end,
]) { ... }
```

And then the body of the function would be identical to the current version, except we wouldn't need `var key = keyOf(value)` line.

A real-world use case for this is in [`dart_ci`][1]'s test results backend, which slices a `List<Result>` sorted by a `String name` field, using a string value provided by the user.

Since we can't search for a `String` in a `List<Result>` using `package:collection` today, the code currently generates `Result`s with just the name field: https://github.com/dart-lang/dart_ci/blob/f3f8e9a1e488c20f559234bff37815fecb94a623/current_results/lib/src/result.dart#L43-L44

In https://dart-review.googlesource.com/c/dart_ci/+/432080 I'm fixing this by implementing my own lower bound function. Ideally though it would be great if I could just use `package:collection`.

[1]: https://github.com/dart-lang/dart_ci

xlkv:
Fixes #895

Previously, `binarySearchBy` and `lowerBoundBy` required passing a full
list element of type `E`, which forced callers to construct a dummy object
just to search by a key field. For example, searching a `List<Result>`
sorted by a `String name` field required creating a `Result` with only
the name populated.

Now both functions accept a key of type `K` directly, matching the type
returned by the `keyOf` function. This enables natural use cases like:

```dart
binarySearchBy(results, (r) => r.name, (a, b) => a.compareTo(b), 'foo');

### DIFF
--- a/pkgs/collection/CHANGELOG.md
+++ b/pkgs/collection/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 1.20.0-wip
 
+- Generalize `binarySearchBy` and `lowerBoundBy` to accept a key of type `K`
+  directly instead of a list element of type `E`. This allows searching a
+  `List<E>` sorted by a key field using a standalone key value without
+  constructing a full `E` object.
+
 - Adds `separated` and `separatedList` extension methods to `Iterable`.
 - Adds `separate` extension method to `List`
 - Add `IterableMapEntryExtension` for working on `Map` as a list of pairs, using
--- a/pkgs/collection/lib/src/algorithms.dart
+++ b/pkgs/collection/lib/src/algorithms.dart
@@ -27,12 +27,12 @@ int binarySearch<E>(
   return binarySearchBy<E, E>(sortedList, identity, compare, value);
 }
 
-/// Returns a position of the [value] in [sortedList], if it is there.
+/// Returns a position of the [key] in [sortedList], if it is there.
 ///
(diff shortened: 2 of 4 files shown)


natebosch:
Hmm, I may have been mistaken to mark that issue as ready for contributions since I was overlooking the breaking change.

@lrhn do you think this is something we have the flexibility to roll in with the major version bump?

lrhn:
It's a good change, but it _is_ breaking, so probably not something we can just do.

Also no good way to make the key or element optional (since the function already has optional positional parameters).
Best solution would probably be to introduce a second function, `binarySearchByKey` taking the key, which the existing function can forward to with `key(element)` as key.

Then deprecate the old one and hope we can make it go away eventually.

PR #960, opened 2026-05-07
WIP: Trying out excerpts

kevmoo:


kevmoo:
### 1. `collection`: Non-functional Legacied `MapEquality` Constructor
* **The Legacy Code in `README`:**
  ```dart
  const MapEquality(IdentityEquality(), ListEquality());
  ```
* **The Bug:** 
  This is a **compile-time error**! The modern `MapEquality` constructor expects **named parameters** (`keys:` and `values:`), not positional arguments. Furthermore, strict generic typing requires type arguments for the child equalities to compile under modern analysis:
  ```dart
  const MapEquality<String, List<int>>(
    keys: IdentityEquality<String>(),
    values: ListEquality<int>(),
  );
  ```
* **Impact:** Anyone copying the legacy snippet directly from the documentation into a modern Dart codebase would be met with a compile error: `Too many positional arguments: 0 expected, but 2 found.`

---

### 2. `logging`: `FutureOr` Return Type Mismatch in `catchError`
* **The Legacy Code in `README`:**
  ```dart
  var future = doSomethingAsync().then((result) {
    log.fine('Got the result: $result');
    processResult(result);
  }).catchError((e, stackTrace) => log.severe('Oh noes!', e, stackTrace));
  ```
* **The Bug:**
  In `dart:async`, `catchError` expects a callback that returns a `FutureOr` of the generic type argument. Because the `doSomethingAsync().then(...)` block returns `void`, the catch block also must return `FutureOr<void>` or `FutureOr<Null>`. 
  However, `Logger.severe()` is a `void` method itself, and in Dart a lambda expression `=> void_returning_expression` evaluates to `void`. The analyzer raises a strict warning or error:
  ```
  A value of type 'void' can't be returned by the 'onError' handler because it must be assignable to 'FutureOr<Null>'.
  ```
  To make this correct, it should be written as an explicit block instead of an arrow function expression, or handle the error return types correctly:
  ```dart
  var future = doSomethingAsync().then((result) {
    log.fine('Got the result: $result');
    processResult(result);
  }).catchError((Object e, StackTrace stackTrace) {
    log.severe('Oh noes!', e, stackTrace);
  });
  ```

---

### 3. `args`: Broken Command Runner Snippet Formatting
* **The Legacy Code in `README`:**
  In the `CommitCommand.run()` snippet:
  ```dart
  void run() {
  ```
  The backticks were broken and split across a code segment block midway through:
  ```markdown
  void run() {
  ```
  `// [argResults] is set before [run()] is called...`
  ```
* **The Bug:** This caused the generated markdown rendering to break completely in formatting on some platforms because of split nested backticks mid-method. We combined it into a single unified code block.

---

### 4. Strict Type Safety & Modern Nullability
* **The Legacy Code in `README`:**
  Across many `README` files, types in callbacks and generic arguments were entirely omitted (e.g., `(e, stackTrace)` in `catchError`).
* **The Bug:**
  Under strict type safety options, untyped error callback parameters implicitly infer to `dynamic`, triggering compilation warnings like `inference_failure_on_untyped_parameter`. Explicitly typing them as `(Object e, StackTrace stackTrace)` is now required for clean, warning-free compilation in new projects.

pkgs/args/README.md
```diff
@@ -11,23 +11,31 @@ server-side and client-side apps.
 
 First create an [ArgParser][]:
 
-    var parser = ArgParser();
+<?code-excerpt "example/readme_examples.dart (parser-init)"?>
```
lrhn, line 14: Is dartdoc _officially specified_ to ignore HTML processing instructions? 

We shouldn't be depending on accidental behavior.

(And I'd still prefer to use a dartdoc example directive over yet-another-tool... But I also like that that code is actually on the source, because I read source. So maybe dartdoc example directives should allow code to be inline as well, and then check that it matches the external region. @jonasfj )

pkgs/args/README.md
```diff
@@ -90,26 +103,32 @@ parser.addOption('mode', mandatory: true);
...
 var results = parser.parse(['some', 'command', 'line', 'args']);
 ```
 
 These arguments usually come from the arguments to `main()`. For example:
 
-    main(List<String> args) {
-      // ...
-      var results = parser.parse(args);
-    }
+<?code-excerpt "example/readme_examples.dart (parser-main-args)" replace="/void mainExample/main/g; /  final parser = ArgParser\(\);\n//g;"?>
```
lrhn, line 113: Replacement functionality feels to miss the point. 

The point of this feature is to check that the example code is being tested, by making sure it's the same as the code in the file that is being tested.

If you can replace arbitrary text using something as ungrammatical as a RegExp, then you've checked nothing. 
If you could only _rename a declaration_, then maybe it's proof of something.
Of if you could only rename a `main` function, maybe that would still be enough.

(In this case it should not remove the `void`!)

pkgs/args/README.md
```diff
@@ -203,6 +228,7 @@ print(results.multiOption('mode')); // prints '[on, off]'
 
 By default, values for a multi-valued option may also be separated with commas:
 
+<?code-excerpt "example/readme_examples.dart (parser-multi-option-commas)" replace="/multiCommaParser/parser/g; /multiCommaResults/results/g;"?>
```
lrhn, line 231: This rename should not be necessary. Just rename in the original.

pkgs/args/README.md
```diff
@@ -251,16 +280,18 @@ You can add multiple commands to the same parser so that a user can select one
 from a range of possible commands. When parsing an argument list, you can then
 determine which command was entered and what options were provided for it.
 
+<?code-excerpt "example/readme_examples.dart (parser-command-results)" replace="/commandResults/results/g; /\?.name/.name/g; /\?\['all'\]/['all']/g;"?>
```
lrhn, line 283: Same, rename should not be needed.

Also removing `?` from `?.` changes meaning significantly. I wouldn't assume the result is valid. Might as well just not use excerpts then. 

(And the `.` should be escaped in the RegExp.
If the example had also contained `test ? name : other`, it would convert that to `test .name : other`. 
If also put a `\b` after `name` do it doesn't match longer names.
People can't write RegExps correctly, having them anywhere near validation makes the outcome much more questionable. Replacements would be safer if they just replaced literal text, and didn't convert it to a RegExp first. And then, again, only entire identifiers which are declared variables, not, fx,  text inside strings )

pkgs/collection/example/readme_examples.dart
```diff
@@ -0,0 +1,16 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// This file contains the code snippets for collection/README.md.
+
+import 'package:collection/collection.dart';
+
+void main() {
+  // #docregion map-equality
+  const MapEquality<String, List<int>>(
+    keys: IdentityEquality<String>(),
```
lrhn, line 12: I would not add the type arguments. 

You don't need them if you don't have the strict dynamic "lint" enabled. It's not in the recommended set.

The classes are designed so that using no type arguments gives you a useful equality. It's what I would recommend, and it should work.

(We should perhaps change the bounds of the types to an explicit `Object?` I avoid strict dynamic issues.)

We should not necessarily test our example code with the same lintsx we use for our own code. If y use the recommended set, and not more than that. Example code is written to be used by people using the recommended lints.

pkgs/crypto/README.md
```diff
@@ -24,18 +24,14 @@ To hash a list of bytes, invoke the [`convert`][convert] method on the
 [`sha1`][sha1-obj], [`sha256`][sha256-obj] or [`md5`][md5-obj]
 objects.
 
+<?code-excerpt "example/readme_examples.dart (digest-single)"?>
 ```dart
-import 'package:crypto/crypto.dart';
-import 'dart:convert'; // for the utf8.encode method
+var bytes = utf8.encode("foobar"); // data being hashed
```
lrhn, line 29: Firmaet comments as statements (says the style guide).

`// Data being hashed.`

(It's OK to just comment what a print prints, but this is a phrase.)

pkgs/crypto/README.md
```diff
@@ -49,26 +45,20 @@ method for each chunk of input data, and invoke the `close` method
 when all the chunks have been added. The digest can then be retrieved
 from the `Sink<Digest>` used to create the input data sink.
 
+<?code-excerpt "example/readme_examples.dart (digest-chunked)" replace="/chunkedDigest/digest/g;"?>
```
lrhn, line 48: Again, replace in the source, not here.

pkgs/os_detect/README.md
```diff
@@ -14,6 +14,7 @@ getters like `isLinux`, `isAndroid` and `isBrowser` based on the
 
 To use this package instead of `dart:io`, replace the import of `dart:io` with:
 
+<?code-excerpt "example/readme_examples.dart (import)" replace="/\/\/ ignore: unused_import\n//g;"?>
```
lrhn, line 17: Can't you put the ignore outside the region?
If you can't, maybe is a dartdoc issue.


Are we sure the line ending won't be `\r\n` on some platforms?

pkgs/os_detect/example/readme_examples.dart
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// This file contains the code snippets for os_detect/README.md.
+
+// #docregion import
+// ignore: unused_import
```
lrhn, line 8: Or just, like, use it!
Or use `ignore-for-file`.

pkgs/path/README.md
```diff
@@ -38,6 +40,7 @@ This calls the top-level `join()` function to join the "directory" and
 If you want to work with paths for a specific platform regardless of the underlying platform that the program is running on, you can create a
 [Context] and give it an explicit [Style]:
 
+<?code-excerpt "example/readme_examples.dart (context)" replace="/p.Style/Style/g;"?>
```
lrhn, line 43: Import or without prefix too, and don't have the prefix in the source.

Or keep the prefix. Every example of an import has `as p`.

(And the RegExp should escape the `.`.)

Can you use two excerpts for over example, so you could except both an import and a later function?

lrhn (commented):
The formatting fixed are nice. Having the code being guaranteed to parse is nice. 

There is no use of renaming that I don't find completely misplaced, and the majority of the RegExps are questionable.
I don't think RegExp-replace is a good feature to have. It encourages not handling the actual problems, just "fixing it in post-production", which is counter to the entire point of having the code in a døDart file. The example code can still be completely wrong after a replace.
That would go for literal text replacing too, being RegExp just makes it even more error prone. (Now you have two problems!)

I think one should be very explicit about which lints (including "modes") are enabled for the example code. It should be the lints of the reader, not the author, which would mean just the recommended lints. 

Enforcing more than that may make you write thighs in the examples that users don't need. 
(It may not matter, but if it does... It probably shouldn't.)

PR #969, opened 2026-07-09
Add `value` getter to `Result`

natebosch:
This is the synchronous version of `asFuture`. Typical uses of `Result`
are asynchronous and the `asFuture` API is sufficient.

The `ValueResult` has a `result` member already, graduating this to the
`Result` super class allows callers to use it without checking the type
first.


natebosch:
We could also write this as an extension getter easily, but I think that these classes were not designed to be extended and it may be worth the potentially breaking change to mark them `final`. If something comes up in internal usage we may want to rethink this plan, but hopefully no one will have used these as mocks or anything like that.

https://github.com/dart-lang/core/pull/970

lrhn:
I'd consider just using the `value` getter, and adding it on `Result` and `ErrorResult`. 
Otherwise there'll be two getters with the same behavior on`ValueResult`.

It does mean you can use `value` without first checking that there is one, without getting a warning through the name mentioning throwing. 
The `valueOrRethrow` is a long and ugly name. If that's intended to discourage use, it might work, but why add a member _and_ discourage it's use. It's easier to not add it. There must be intended uses, and making those inconvenient is just annoying.

I'd reuse `value`. WDYT?

natebosch:
> I'd reuse `value`. WDYT?

SGTM. I tried it out and it reads well to me in practice.

pkgs/async/test/result/result_test.dart
```diff
@@ -131,6 +131,29 @@ void main() {
...
+
+    test('throws error for ErrorResult', () {
+      var result = Result<int>.error('BAD', stack);
+      expect(() => result.value, throwsA('BAD'));
+    });
+
+    test('throws error with stack trace for ErrorResult', () {
+      var result = Result<int>.error('BAD', stack);
+      try {
+        result.value;
```
lrhn, line 149: Can't it use `expect(…, throwsA(…))`?

If not, maybe document why.

lrhn (approved):

lrhn:
Also change commit title

pkgs/async/test/result/result_test.dart
natebosch: We don't have a good pattern for testing expectations on stack traaces.

I filed https://github.com/dart-lang/test/issues/2706 to potentially add a better way in `package:checks`. If we end up using a getter based model instead of `Condition` arguments I think using a `Subject<({Object error, StackTrace stackTrace})>` might be a good choice

PR #970, opened 2026-07-09
Make `Result` sealed and subclasses final

natebosch:
These were never intended for extension and marking them `final` makes
it easier to expand the API without breaking changes.

- Mark `Result` as `sealed class`.
- Mark `ValueResult` and `ErrorResult` as `final class`.
- Refactor `Result`, `ValueResult`, and `ErrorResult` into a single
  library using `part` and `part of` to allow inheritance within the
  library.
- Remove redundant exports of `error.dart` and `value.dart` from
  `lib/async.dart`.


natebosch:
This passed an internal test run, I think it's worth landing.

pkgs/async/lib/src/result/result.dart
```diff
@@ -22,7 +23,7 @@ import 'value.dart';
 ///
 /// A [Future] represents a potential result, one that might not have been
 /// computed yet, and a [Result] is always a completed and available result.
-abstract class Result<T> {
+abstract final class Result<T> {
```
lrhn: Could it be sealed?

lrhn (approved):

pkgs/async/lib/src/result/result.dart
natebosch: Yes, good call.

lrhn (approved):

PR #975, opened 2026-07-10
Enable changelog updates in groundskeeper workflow

mosuem:
Ported changes from labs groundskeeper workflow to enable changelog updates with a default message.


natebosch:
I tend to omit stuff from the changelog when it's not user facing. I think we expect automated changes like that to have no impact on end users, and it's better to optimize for that case than to add changelog entries in case an automated fix happens to be unexpectedly breaking.

mosuem:
Ok, sounds good to me.

PR #979, opened 2026-07-17
Port `pairs` extension from dart-archive/collection#289

samlythemanly:
This ports the `pairs` extension from dart-archive/collection#289.

The original implementation was approved but got derailed by rollup attempts that included a major version bump and breaking changes (dart-archive/collection#284 and dart-lang/core#859). Those PRs have been in limbo for years and are unlikely to ever be merged.

---

<details>

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

### DIFF
--- a/pkgs/collection/lib/src/map_extensions.dart
+++ b/pkgs/collection/lib/src/map_extensions.dart
@@ -1,0 +1,8 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+extension MapExtensions<K, V> on Map<K, V> {
+  /// Like [Map.entries], but returns each entry as a record instead of a [MapEntry].
+  Iterable<(K, V)> get pairs => entries.map((entry) => (entry.key, entry.value));
+}
--- a/pkgs/collection/test/extensions_test.dart
+++ b/pkgs/collection/test/extensions_test.dart
@@ -2393,6 +2393,19 @@ void main() {
       });
     });
   });
+
+  group('Map', () {
+    group('pairs', () {
+      test('returns an empty list for an empty map', () {
+        expect(const <int, int>{}.pairs, isEmpty);
+      });
+
+      test('returns a pair for every entry in the map', () {
+        expect(const {1: 2, 3: 4, 5: 6}.pairs,
+            equals(const [(1, 2), (3, 4), (5, 6)]));
+      });
+    });
+  });
 }
 
 /// Creates a plain iterable not implementing any other class.
--- a/pkgs/collection/CHANGELOG.md
+++ b/pkgs/collection/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 1.20.0-wip
 
+- Adds `Map.pairs` extension method to make it easier to work with maps using records rather than MapEntry.
 - Adds `separated` and `separatedList` extension methods to `Iterable`.
 - Adds `separate` extension method to `List`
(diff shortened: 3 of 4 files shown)


mosuem:
@dart-lang/dart-core-packages-team any feedback here?

pkgs/collection/lib/src/map_extensions.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+extension MapExtensions<K, V> on Map<K, V> {
+  /// Like [Map.entries], but returns each entry as a record instead of a [MapEntry].
```
lrhn, line 6: Docs should say what it is from scratch, not start out relative to something else. 
Iterables don't "return" anything. 

Consider: 
> ```
> /// The key and value pairs of this map. 
> ///
> /// Contains `(key, value)` for each key `key`
> /// and it's associated value, in key iteration
> /// order.
> ```

Should probably also document that `contains` uses equality of record pairs, it does not use the map's lookup, and it's not efficient.

(The `.map` iterable should have efficient length, but not much else.)

pkgs/collection/test/extensions_test.dart
```diff
@@ -2393,6 +2393,19 @@ void main() {
...
+      test('returns an empty list for an empty map', () {
+        expect(const <int, int>{}.pairs, isEmpty);
+      });
+
+      test('returns a pair for every entry in the map', () {
+        expect(const {1: 2, 3: 4, 5: 6}.pairs,
+            equals(const [(1, 2), (3, 4), (5, 6)]));
+      });
+    });
+  });
```
lrhn, line 2408: A little sparsely tested.
Try with maps that don't use normal equality. 
Maybe an identity map with two equal-but-district keys.

lrhn (approved):
It's a perfectly usable implementation. 

I'm still not sold on having the `pairs` getter, but it shouldn't hurt (other than wastefully allocating the `MapEntry` objects).

pkgs/collection/lib/src/map_extensions.dart
rakudrama: `Map.entries` is silent on the order, so can we promise key iteration order?  https://api.dart.dev/dart-core/Map/entries.html — perhaps the unspecified order is an oversight — the default implementation iterates the keys https://api.dart.dev/dart-collection/MapBase/entries.html

arthurbcd: Shouldn't we base this on the default implementation, then? By doing so, we avoid allocating `MapEntry`.

```dart
Iterable<(K, V)> get pairs {
  return keys.map((K key) => (key, this[key] as V));
}
```

https://api.dart.dev/dart-collection/MapBase/entries.html

lrhn: The `MapBase` is not the _default_ implantation, it's a _fallback_ implementation using only limited primitives. Any `Map` that cares about performance should have a better version.

(Maps should have had an `Iterable<R> mapEntries<R>(R Function(K, V) convert)`. Then you can create your own by iterables without allocating entry objects. 
Alas, not so.

The entry objects should be short lived, so cheap for a generational GC.

And the unspecified order it's a mistake, it is always in key iteration order. The `Map` class documentation speaks about iteration order in general, because it differs between maps, but `keys`, `values` and `entries` should use the same order.

PR #981, opened 2026-07-23
[path] Strip Windows Extended-Length Path prefixes in absolutePathToUri

### LINKED ISSUE #980: [path] \WindowsStyle.absolutePathToUri\ throws FormatException on Windows Extended-Length UNC paths
### Summary

`package:path`'s `WindowsStyle.absolutePathToUri` throws `FormatException` when given a Windows Extended-Length UNC path of the form `\\?\UNC\server\share\...`. These paths are returned by `Directory.resolveSymbolicLinksSync()` (which calls the Windows API `GetFinalPathNameByHandle`) for any UNC path, and are also the documented way to bypass MAX_PATH limits on Windows.

Note: this is **not** the root cause of Analysis Server failing on WSL projects (that's [analyzer's `_PhysicalResource.exists` missing try-catch](https://github.com/dart-lang/sdk), filed separately). But it is a real bug that affects any code calling `path.toUri()` on a `resolveSymbolicLinksSync()` result.

### Reproduction

`pubspec.yaml`:
```yaml
name: test_path_unc
environment:
  sdk: ^3.0.0
dependencies:
  path: ^1.9.0
```

`test.dart`:
```dart
import 'dart:io';
import 'package:path/path.dart' as p;

void main() {
  final resolved = Directory.current.resolveSymbolicLinksSync();
  print('resolved: $resolved');

  final uri = p.toUri(resolved);   // <- throws
  print('uri: $uri');
}
```

Run from a UNC/mapped-drive directory:
```
cd W:\tmp\test_path_unc   (W: -> \\wsl.localhost\Ubuntu-24.04)
dart run test.dart
```

### Actual output

```
resolved: \\?\UNC\wsl.localhost\Ubuntu-24.04\tmp\test_path_unc

FormatException: Invalid character (at character 1)
?
^

#0  _Uri._fail (dart:core/uri.dart:2050:5)
#1  _Uri._normalizeRegName (dart:core/uri.dart:2598:9)
#2  _Uri._makeHost (dart:core/uri.dart:2453:12)
#3  new _Uri (dart:core/uri.dart:1955:12)
#4  WindowsStyle.absolutePathToUri (package:path/src/style/windows.dart:121:14)
#5  Context.toUri (package:path/src/context.dart:1062:20)
#6  toUri (package:path/path.dart:459:35)
```

### Boundary test

| Input | Result |
|-------|--------|
| `\\?\UNC\wsl.localhost\Ubuntu-24.04\home\berial` | FAIL |
| `\\?\UNC\localhost\c$\Users` | FAIL |
| `\\?\UNC\127.0.0.1\c$\Users` | FAIL |
| `\\?\UNC\someserver\share\path` | FAIL (even non-existent server) |
| `\\wsl.localhost\Ubuntu-24.04\home\berial` | OK |
| `\\localhost\c$\Users` | OK |
| `\\someserver\share\path` | OK |
| `W:\home\berial` | OK |

So the bug is not WSL-specific — it affects **all** Windows Extended UNC paths (`\\?\UNC\*`).

Note: `Uri.file()` (from `dart:core`) handles `\\?\UNC\...` correctly (strips the prefix), so this is specifically a `package:path` issue, not a `dart:core` issue.

### Root cause

[`pkgs/path/lib/src/style/windows.dart:29`](https://github.com/dart-lang/core/blob/main/pkgs/path/lib/src/style/windows.dart#L29):

```dart
final rootPattern = RegExp(r'^(\\\\[^\\]+\\[^\\/]+|[a-zA-Z]:[/\\])');
```

The first alternative `\\\\[^\\]+\\[^\\/]+` is meant to match `\\server\share`, but it also matches `\\?\UNC` because `[^\\]+` happily matches `?`.

Then in `absolutePathToUri` ([line 121](https://github.com/dart-lang/core/blob/main/pkgs/path/lib/src/style/windows.dart#L121)):

```dart
final rootParts = parsed.root!.split('\\').where((part) => part != '');
parsed.parts.insert(0, rootParts.last);
// ...
return Uri(scheme: 'file', host: rootParts.first, pathSegments: parsed.parts);
```

- `parsed.root` = `\\?\UNC`
- `split('\\')` → `['', '', '?', 'UNC']`
- `where((part) => part != '')` → `['?', 'UNC']`
- `rootParts.first` = `'?'`
- `Uri(host: '?')` → FormatException (`?` is not a valid URI host)

### Suggested fix

Either strip the `\\?\` / `\\?\UNC\` prefix before parsing, or refine the `rootPattern` to not match `\\?\`:

```dart
// Option A: strip extended-length prefix in absolutePathToUri
Uri absolutePathToUri(String path) {
  if (path.startsWith(r'\\?\UNC\\')) {
    path = r'\\' + path.substring(8);
  } else if (path.startsWith(r'\\?\')) {
    path = path.substring(4);
  }
  // ... existing logic
}
```

```dart
// Option B: refine rootPattern
final rootPattern = RegExp(r'^(\\\\[^\\?][^\\]*\\[^\\/]+|[a-zA-Z]:[/\\])');
```

### Environment

- Dart SDK: 3.11.5 (stable) on `windows_x64`
- package:path: 1.9.1 (also confirmed `main` branch is affected)
- OS: Windows 11 + WSL2 (Ubuntu-24.04)

berial:
## Problem

`WindowsStyle.absolutePathToUri` (and therefore `toUri()`) throws `FormatException` when given Windows Extended-Length Path prefixes:

- `\\?\C:\long\path`
- `\\?\UNC\server\share\path`

These prefixes are documented in [Win32 File Namespaces](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces) and are increasingly common because:

1. `GetFinalPathNameByHandle` always returns the `\\?\UNC\` form for any UNC path
2. WSL paths (`\\wsl.localhost\<distro>\...`) hit this format whenever Windows APIs normalize them
3. Paths exceeding `MAX_PATH` (260 chars) require the `\\?\` prefix

### Repro

```dart
import 'package:path/path.dart' as p;

void main() {
  // Works fine — Uri.file() strips the prefix
  print(Uri.file(r'\\?\UNC\wsl.localhost\Ubuntu-24.04\tmp\foo'));
  // → file://wsl.localhost/Ubuntu-24.04/tmp/foo

  // Throws FormatException — path.toUri() does not strip the prefix
  print(p.toUri(r'\\?\UNC\wsl.localhost\Ubuntu-24.04\tmp\foo'));
  // → FormatException: Illegal character in path
}
```

### Root cause

`absolutePathToUri` does not strip the `\\?\` prefix before parsing. `ParsedPath.parse` then misinterprets `\\?\UNC` as a UNC root with `?` as the server name, which is not a valid URI host — causing `FormatException` when passed to the `Uri` constructor.

This makes `path.toUri()` inconsistent with `Uri.file()` / `Uri.directory()` from `dart:io`, which already handle these prefixes correctly.

## Fix

Strip the Win32 Extended-Length Path prefix at the entry of `absolutePathToUri`, before any parsing:

- `\\?\UNC\server\share\...` → `\\server\share\...` (strip `\\?\UNC\`, prepend `\\`)
- `\\?\C:\path\...` → `C:\path\...` (strip `\\?\`)

This aligns `path.toUri()` with `Uri.file()` — the `\\?\` prefix is a Win32 API artifact (enables paths > 260 chars and disables DOS device-name interpretation), not user intent, and should not leak into the resulting file URI.

Discussed in #980 — @lrhn confirmed the direction: *"Maybe we just need to see what `Uri.fromFile` does, and copy that."*

## Tests

Added 6 test cases to `pkgs/path/test/windows_test.dart` covering:

| Input | Expected URI |
|-------|-------------|
| `\\?\UNC\server\share\path\to\foo` | `file://server/share/path/to/foo` |
| `\\?\UNC\wsl.localhost\Ubuntu-24.04\tmp\foo` | `file://wsl.localhost/Ubuntu-24.04/tmp/foo` |
| `\\?\UNC\server\share` | `file://server/share` |
| `\\?\UNC\server\share\` | `file://server/share/` |
| `\\?\C:\path\to\foo` | `file:///C:/path/to/foo` |
| `\\?\C:\` | `file:///C:/` |

All existing tests + 6 new tests pass (68 total in `windows_test.dart`). `dart analyze` reports no issues.

## Scope

This PR intentionally only fixes `absolutePathToUri` (the function reported in #980). Other methods like `rootLength` / `split` may also benefit from prefix stripping, but those are out of scope for this focused fix.

Fixes #980

### DIFF
--- a/pkgs/path/lib/src/style/windows.dart
+++ b/pkgs/path/lib/src/style/windows.dart
@@ -103,6 +103,24 @@ class WindowsStyle extends InternalStyle {
 
   @override
   Uri absolutePathToUri(String path) {
+    // Strip Windows Extended-Length Path prefixes so that `toUri()` aligns
+    // with `Uri.file()` / `Uri.directory()` from `dart:io`.
+    //
+    // `\\?\C:\path`           -> `C:\path`
+    // `\\?\UNC\server\share` -> `\\server\share`
+    //
+    // The `\\?\` prefix is a Win32 API artifact (enables paths > 260 chars
+    // and disables DOS device-name interpretation). It is not user intent and
+    // should not leak into the resulting file URI. See:
+    // https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces
+    if (path.startsWith(r'\\?\')) {
+      if (path.startsWith(r'\\?\UNC\')) {
+        path = r'\\' + path.substring(8);
+      } else {
+        path = path.substring(4);
+      }
+    }
+
     final parsed = ParsedPath.parse(path, this);
     if (parsed.root!.startsWith(r'\\')) {
       // Network paths become "file://server/share/path/to/file".
@@ -119,10 +137,7 @@ class WindowsStyle extends InternalStyle {
       }
 
       return Uri(
-        scheme: 'file',
-        host: rootParts.first,
-        pathSegments: parsed.parts,
-      );
+          scheme: 'file', host: rootParts.first, pathSegments: parsed.parts);
     } else {
       // Drive-letter paths become "file:///C:/path/to/file".
 
@@ -136,10 +151,8 @@ class WindowsStyle extends InternalStyle {
 
       // Get rid of the trailing "\" in "C:\" because the URI constructor will
       // add a separator on its own.
-      parsed.parts.insert(
-        0,
-        parsed.root!.replaceAll('/', '').replaceAll('\\', ''),
-      );
+      parsed.parts
+          .insert(0, parsed.root!.replaceAll('/', '').replaceAll('\\', ''));
 
       return Uri(scheme: 'file', pathSegments: parsed.parts);
     }
--- a/pkgs/path/lib/src/style/windows.dart
+++ b/pkgs/path/lib/src/style/windows.dart
@@ -114,10 +114,10 @@ class WindowsStyle extends InternalStyle {
     // should not leak into the resulting file URI. See:
     // https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces
     if (path.startsWith(r'\\?\')) {
-      if (path.startsWith(r'\\?\UNC\')) {
-        path = r'\\' + path.substring(8);
+      if (path.startsWith(r'UNC\', r'\\?\'.length)) {
+        path = path.replaceRange(0, r'\\?\UNC'.length, r'\');
       } else {
-        path = path.substring(4);
+        path = path.substring(r'\\?\'.length);
       }
     }
 
--- a/pkgs/path/CHANGELOG.md
+++ b/pkgs/path/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 1.9.2-wip
 
+- `WindowsStyle.absolutePathToUri` (and therefore `toUri()`) now strips Windows
+  Extended-Length Path prefixes (`\\?\` and `\\?\UNC\`) instead of
+  throwing a `FormatException`. This aligns `path.toUri()` with
+  `Uri.file()` / `Uri.directory()` from `dart:io`.
 - Make `Url` style better at recognizing schemes and authorities.
   Only consider a path as being a schemed URL if it starts with a
   valid scheme.
--- a/pkgs/path/test/windows_test.dart
+++ b/pkgs/path/test/windows_test.dart
@@ -1058,6 +1058,31 @@ void main() {
       Uri.parse('_%7B_%7D_%60_%5E_%20_%22_%25_'),
     );
     expect(context.toUri(''), Uri.parse(''));
+
+    // Windows Extended-Length Path prefixes are stripped to align with
+    // `Uri.file()` from `dart:io`. See:
+    // https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces
+    expect(
+      context.toUri(r'\\?\UNC\server\share\path\to\foo'),
+      Uri.parse('file://server/share/path/to/foo'),
+    );
+    expect(
+      context.toUri(r'\\?\UNC\wsl.localhost\Ubuntu-24.04\tmp\foo'),
+      Uri.parse('file://wsl.localhost/Ubuntu-24.04/tmp/foo'),
+    );
+    expect(
+      context.toUri(r'\\?\UNC\server\share'),
+      Uri.parse('file://server/share'),
+    );
+    expect(
+      context.toUri(r'\\?\UNC\server\share\'),
+      Uri.parse('file://server/share/'),
+    );
+    expect(
+      context.toUri(r'\\?\C:\path\to\foo'),
+      Uri.parse('file:///C:/path/to/foo'),
+    );
+    expect(context.toUri(r'\\?\C:\'), Uri.parse('file:///C:/'));
   });
 
   group('prettyUri', () {


pkgs/path/lib/src/style/windows.dart
```diff
@@ -103,6 +103,24 @@ class WindowsStyle extends InternalStyle {
...
+    //
+    // `\\?\C:\path`           -> `C:\path`
+    // `\\?\UNC\server\share` -> `\\server\share`
+    //
+    // The `\\?\` prefix is a Win32 API artifact (enables paths > 260 chars
+    // and disables DOS device-name interpretation). It is not user intent and
+    // should not leak into the resulting file URI. See:
+    // https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces
+    if (path.startsWith(r'\\?\')) {
+      if (path.startsWith(r'\\?\UNC\')) {
```
lrhn: Can be `(path.startsWith(r'UNC\', r'\\?\'.length))` to not check the same chars again.

pkgs/path/lib/src/style/windows.dart
```diff
@@ -103,6 +103,24 @@ class WindowsStyle extends InternalStyle {
...
+        path = r'\\' + path.substring(8);
```
lrhn: Can be `path = r'\' + path.substring(7);`. 
I'd probably do `path = path.replaceRange(0, r'\\?\UNC'.length, r'\');`.
(Creates fewer intermediate strings.)

pkgs/path/lib/src/style/windows.dart
```diff
@@ -103,6 +103,24 @@ class WindowsStyle extends InternalStyle {
...
+      } else {
+        path = path.substring(4);
```
lrhn: `path = path.substring(r'\\?\'.length);`

lrhn: This doesn't handle `\\?\PIPE\...`. We can probably assume that won't happen for any reasonable use, but I think it is a valid Win32 path. (Not entirely sure for _what_, probably named pipes. With more than 260 chars in the name!)

That is one argument for not necessarily considering `\\?\UNC\` as unintended by the user. Maybe they know what they're doing. (But the common user don't, so maybe we just need a backdoor to allow preserving the path.)

pkgs/path/lib/src/style/windows.dart
```diff
@@ -103,6 +103,24 @@ class WindowsStyle extends InternalStyle {
 
   @override
   Uri absolutePathToUri(String path) {
+    // Strip Windows Extended-Length Path prefixes so that `toUri()` aligns
+    // with `Uri.file()` / `Uri.directory()` from `dart:io`.
+    //
+    // `\\?\C:\path`           -> `C:\path`
```
lrhn, line 109: The spacing won't be visible in rendered markdown, so might as well remove it.
Consider making it an itemized list, ``* `\\?\C:\path` ...``.

You can use `&rarr;` or `&mapsto;` for a nicer-looking arrow.

pkgs/path/lib/src/style/windows.dart
```diff
@@ -103,6 +103,24 @@ class WindowsStyle extends InternalStyle {
...
+    // `\\?\UNC\server\share` -> `\\server\share`
+    //
+    // The `\\?\` prefix is a Win32 API artifact (enables paths > 260 chars
+    // and disables DOS device-name interpretation). It is not user intent and
+    // should not leak into the resulting file URI. See:
```
lrhn, line 114: (Are we absolutely sure we put the `\\?\` back in front of paths containing device names? Otherwise removing it from `\\?\C:\nul.txt` will make the path not point to the same file again.
Do we have tests?)

pkgs/path/lib/src/style/windows.dart
berial: Agreed — `\\?\PIPE\...` is a Win32 named pipe, which lives in the device namespace, not the filesystem. Converting it to a `file:` URI is semantically meaningless since `file:` URIs only address filesystem resources. (Dart accesses named pipes via `Socket`/IPC, not `File`.)

With the current fix, `\\?\PIPE\foo` becomes `\PIPE\foo` after stripping the `\\?\` prefix. Since a single leading backslash is not a UNC root, `ParsedPath.parse` treats it as a relative path — no throw, but also not a "real" pipe path. That's a degenerate case I think we can accept, since there's no correct `file:` URI representation to produce anyway.

If a user genuinely needs to preserve the `\\?\` prefix (e.g., to opt back into the >260-char path semantics that `dart:io` handles at the Win32 boundary, as you noted), an opt-in `preserveExtendedPrefix` flag on `toUri`/`absolutePathToUri` would be the cleaner backdoor. But I'd leave that out of this PR unless you want it now — the current change just stops the spurious throw for the common filesystem cases (`\\?\C:\...` and `\\?\UNC\...`), which is what's biting real-world tools.

pkgs/path/lib/src/style/windows.dart
```diff
@@ -103,6 +103,24 @@ class WindowsStyle extends InternalStyle {
...
+    // https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces
+    if (path.startsWith(r'\\?\')) {
+      if (path.startsWith(r'UNC\', r'\\?\'.length)) {
+        path = path.replaceRange(0, r'\\?\UNC'.length, r'\');
```
lrhn, line 118: (Can be `path.replaceRange(1, r'\\?\UNC'.length, '');`. I doubt it makes any difference. It just feels slightly wrong to me to remove a `\` and then add one back again.)

PR #989, opened 2026-08-11
Add new lints

### LINKED ISSUE #954: Consider new lints for inclusion v3.9.0-3.11.0
New lints:
- [simplify_variable_pattern](https://dart.dev/tools/linter-rules/simplify_variable_pattern)
- [remove_deprecations_in_breaking_versions](https://dart.dev/tools/linter-rules/remove_deprecations_in_breaking_versions)
- [switch_on_type](https://dart.dev/tools/linter-rules/switch_on_type)
- [unnecessary_unawaited](https://dart.dev/tools/linter-rules/unnecessary_unawaited)

I think we can start with a single issue to cover all 4 and if initial votes are aligned we can skip detailed discussion. We can fork an issue for anything which isn't aligned to start.

### LINKED ISSUE #978: Consider new lints for inclusion simple_directive_paths
https://dart.dev/tools/linter-rules/simple_directive_paths

PREFER simple directive paths.

Using simple directive paths (avoiding redundant ./ or backtracking ../ segments) makes the code cleaner, more consistent, and easier to read. It also reduces the risk of confusion when resolving paths relative to the source file.

BAD:

```dart
import '../src/bar.dart';
```

GOOD:

```dart
import 'bar.dart';
```

natebosch:
Closes #954
Closes #978

Bump to SDK 3.12.0 to restrict to versions including the new lints.

core:
- switch_on_type

recommended:
- simple_directive_paths
- simplify_variable_pattern
- unnecessary_unawaited

### DIFF
--- a/pkgs/lints/lib/core.yaml
+++ b/pkgs/lints/lib/core.yaml
@@ -34,10 +34,12 @@ linter:
     - provide_deprecation_message
     - secure_pubspec_urls
     - strict_top_level_inference
+    - switch_on_type
     - type_literal_in_constant_pattern
     - unintended_html_in_doc_comment
     - unnecessary_overrides
     - unrelated_type_equality_checks
     - use_string_in_part_of_directives
     - valid_regexps
+    - var_with_no_type_annotation
     - void_checks
--- a/pkgs/lints/lib/core.yaml
+++ b/pkgs/lints/lib/core.yaml
@@ -41,5 +41,4 @@ linter:
     - unrelated_type_equality_checks
     - use_string_in_part_of_directives
     - valid_regexps
-    - var_with_no_type_annotation
     - void_checks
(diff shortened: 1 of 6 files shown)


pkgs/lints/lib/core.yaml
```diff
@@ -34,10 +34,12 @@ linter:
...
     - secure_pubspec_urls
     - strict_top_level_inference
+    - switch_on_type
     - type_literal_in_constant_pattern
     - unintended_html_in_doc_comment
     - unnecessary_overrides
     - unrelated_type_equality_checks
     - use_string_in_part_of_directives
     - valid_regexps
+    - var_with_no_type_annotation
```
lrhn: Is this lint ready to be released? 

It says
> Avoid declaring parameters with `var` and no type annotation.

and

> The analyzer produces this diagnostic when a parameter is declared using the keyword `var` without a type annotation.


You can't declare a variable with `var` _and_ a type annotation, so that should just be "Avoid declaring parameters with `var`".

It says that `var x` gives the parameter the type `dynamic`, but that's not the `var` doing that. You can remove the `var` and just write `x`, and it's just as dynamic. It's not writing a type that makes it dynamic. Unless it inherits a type, and `var` won't prevent that.

The text only mentions `var` as something that will become invalid in future language versions, but it's just as future-unsafe to use `final` (with or without a type).

The lint doesn't sound like it knows what it's doing. It needs at least better documentation, and should preferably also lint on `final`.

lrhn (approved):
LGTM, but I would prefer waiting with `var_with_no_type_annotation` until, at least, it's documentation makes sense.

PR #993, opened 2026-08-19
test(crypto): add C2SP/Wycheproof test vectors for HMAC

kevmoo:
Integrates HMAC test vectors from Project Wycheproof, maintained by the
Community Cryptography Specification Project (C2SP, originally developed
at Google).

### Provenance & Cryptographic Value
Project Wycheproof provides test vectors that specifically stress-test
cryptographic edge cases, boundary conditions, and known attack vectors
rather than just basic happy-path conformance. For HMAC, this includes:
- Key-length boundary transitions: keys shorter than, equal to, and
  longer than hash block sizes (triggering RFC 2104 pre-hashing).
- Edge-case message inputs (zero-length and boundary chunks).
- Tag truncation testing matrices down to sub-byte boundaries.
- Corrupted tags (single-bit flips, all-0/all-1 substitutions).

### Summary of Changes
- Added tool/pull_wycheproof.dart to fetch and pin the HMAC test vector
  JSON suites from C2SP/wycheproof.
- Generated inlined lookup data in test/wycheproof_data.dart with base64
  encoding to enable zero-IO execution on browser (--platform chrome) and
  WebAssembly (--compiler dart2wasm) CI test runners while avoiding false
  positive secret scanner triggers.
- Added test/wycheproof_hmac_test.dart asserting 1,212 test cases across
  SHA-1, SHA-224, SHA-256, SHA-384, SHA-512, SHA-512/224, and SHA-512/256.


pkgs/crypto/tool/pull_wycheproof.dart
```diff
@@ -0,0 +1,89 @@
...
+  'hmac_sha512_256_test.json',
+  'hmac_sha1_test.json',
+];
+
+Future<void> main() async {
+  final client = HttpClient();
+  final vectorContents = <String, String>{};
+
+  for (final file in _hmacVectors) {
+    final uri = Uri.parse('$_baseUrl/$file');
```
natebosch: [nit] Prefer using `Uri` APIs to manipulate URIs. Consider using

```dart
final _baseUrl = Uri.https(
  'raw.githubusercontent.com',
  'C2SP/wycheproof/main/testvectors_v1',
);

//...

    final uri = _baseUrl.resolve(file);
```

pkgs/crypto/.gitignore
```diff
@@ -1,3 +1,4 @@
 .dart_tool
 .packages
 pubspec.lock
+test/data/
```
natebosch: Why do we need this? Is there something writing data here?

`test/data/` is a likely candidate for a meaningful directory in Dart tests. Can we use a different dir if some process is creating local-but-not-transient files?

natebosch (commented):
Is there any licensing we need to worry about?

pkgs/crypto/tool/pull_wycheproof.dart
kevmoo: Done! Updated to use Uri.https and _baseUrl.resolve(file) in tool/pull_wycheproof.dart.

pkgs/crypto/.gitignore
kevmoo: Removed! We initially staged raw JSON files into test/data/wycheproof before inlining into test/wycheproof_data.dart. Since the generator script now creates the inlined data directly in memory, test/data/ is completely unused and has been removed from .gitignore.

kevmoo:
Project Wycheproof is licensed under **Apache 2.0** (originally created and open-sourced by Google security researchers, currently maintained under C2SP).

Because it is a permissive license and only used in `test/` and `tool/` (with zero code entering `lib/`), it is fully compatible with `package:crypto`'s BSD 3-Clause license and does not affect the published package's runtime licensing.

I've added an explicit Apache 2.0 attribution notice in the generated test data and tool headers to satisfy Apache § 4 attribution.

kevmoo:
Friendly ping CC @jonasfj @brianquinlan

pkgs/crypto/tool/pull_wycheproof.dart
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Test vectors are sourced from Project Wycheproof
```
brianquinlan, line 5: Maybe explain what this does? Namely that it generates /test/wycheproof_data.dart

brianquinlan:
@natebosch are you OK with the licensing?

pkgs/crypto/test/wycheproof_data.dart
```diff
@@ -0,0 +1,9524 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
```
natebosch: This feels inappropriate. Can we remove this one?

pkgs/crypto/test/wycheproof_data.dart
```diff
@@ -0,0 +1,9524 @@
...
+
+// Test vectors are sourced from Project Wycheproof (https://github.com/C2SP/wycheproof).
+// Copyright 2017-2026 Google LLC / C2SP.
+// Licensed under the Apache License, Version 2.0.
+
+// GENERATED CODE - DO NOT MODIFY BY HAND
```
natebosch, line 6: Can you add this to the linguist config so it is collapsed by default on github reviews?

pkgs/crypto/tool/pull_wycheproof.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  // Generate test/wycheproof_data.dart with wrapped base64-encoded strings
+  // (chunked at 70 chars) matching the pubviz asset-inlining pattern.
+  final buffer = StringBuffer()
+    ..writeln('// Copyright (c) 2026, the Dart project authors.  '
+        'Please see the AUTHORS file')
+    ..writeln('// for details. All rights reserved. Use of this source code '
+        'is governed by a')
+    ..writeln('// BSD-style license that can be found in the LICENSE file.')
+    ..writeln()
```
natebosch:
```suggestion
```

natebosch (changes requested):

natebosch:
I'm fairly confident we need a full copy of the APACHE license next to this file somewhere, but I'm not confident on the details of where it should be. I _think_ maybe something like a `licenses/LICENSE-wycheproof.txt` file is sufficient if we update the generated file to reference it.

pkgs/crypto/tool/pull_wycheproof.dart
kevmoo: Added a doc comment explaining that this tool fetches HMAC test vectors from Project Wycheproof and generates test/wycheproof_data.dart.

pkgs/crypto/test/wycheproof_data.dart
kevmoo: Done. Removed the Dart BSD license header from the generated file and referenced third_party/wycheproof/LICENSE instead.

pkgs/crypto/test/wycheproof_data.dart
kevmoo: Added .gitattributes marking pkgs/crypto/test/wycheproof_data.dart as linguist-generated=true so GitHub collapses it by default in reviews.

pkgs/crypto/tool/pull_wycheproof.dart
kevmoo: Done. Removed the BSD license header output from the generator script.

kevmoo:
Regarding the license location:

Because `test/` is not excluded in `.gitignore` or `pubspec.yaml`, `test/wycheproof_data.dart` is packaged and distributed in the `package:crypto` archive on pub.dev. To comply with Apache 2.0 § 4(a) (requiring recipients of the Work or Derivative Works to be given a copy of the License), the license file must accompany the distributed package rather than living at the monorepo root (which pub publishing ignores).

Following the existing convention in this monorepo (`pkgs/characters/third_party/Unicode_Consortium/`):
- Placed the full Apache 2.0 license at `pkgs/crypto/third_party/wycheproof/LICENSE`.
- Removed the Dart BSD license header from `wycheproof_data.dart` and `pull_wycheproof.dart`, pointing instead to `third_party/wycheproof/LICENSE`.
- Added `.gitattributes` to mark `pkgs/crypto/test/wycheproof_data.dart` as `linguist-generated=true` so GitHub automatically collapses it in PR diffs.
- Added a doc comment to `pkgs/crypto/tool/pull_wycheproof.dart`.

natebosch (approved):

PR #999, opened 2026-09-10
Update convert package README and add examples

jonasfj:
- Added clear examples to README.md for hex, percent, and fixed datetime.
- Created standalone example files for hex.dart and fixed_datetime_formatter.dart.

### DIFF
--- a/pkgs/convert/README.md
+++ b/pkgs/convert/README.md
@@ -7,3 +7,69 @@ data representations. It's the external counterpart of the
 [`dart:convert`](https://api.dart.dev/dart-convert/dart-convert-library.html)
 SDK library, and contains less-central APIs and APIs that need more flexible
 versioning.
+
+## Hexadecimal
+
+The `hex` codec converts byte arrays to and from hexadecimal strings, following
+[the Base16 spec](https://tools.ietf.org/html/rfc4648#section-8).
+
+```dart
+import 'package:convert/convert.dart';
+
+void main() {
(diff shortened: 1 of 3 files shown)


pkgs/convert/README.md
```diff
@@ -7,3 +7,69 @@ data representations. It's the external counterpart of the
...
+  var decoded = percent.decode(encoded);
+  print(utf8.decode(decoded)); // 'hello world!'
+}
+```
+
+## FixedDateTimeFormatter
+
+A formatter and parser for `DateTime` in a fixed format string pattern.
+This differs from `DateFormat` from `package:intl` in that the format character
+count is interpreted literally.
```
lrhn, line 59: Why use this then?

"Use for converting a `DateTime` to a specific simple output format where you can control the length for each entry. May be faster parsing that same simple format back."

Maybe?

pkgs/convert/example/hex.dart
```diff
@@ -0,0 +1,17 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:convert/convert.dart';
+
+void main() {
+  final bytes = <int>[0, 1, 2, 3, 255];
+
+  // Encode byte array to hex string.
+  final encoded = hex.encode(bytes);
```
lrhn, line 11: (Does it at least decode to `Uint8List`?)

lrhn (approved):

PR #1003, opened 2026-09-14
[os_detect] Deprecate package:os_detect in favor of package:platform

mit-mit:

<details>

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>


pkgs/os_detect/README.md
```diff
@@ -4,6 +4,40 @@
...
+
+foo() {
+  if (os_detect.isAndroid) {
+    ...
+  }
+}
+```
+
+The APIs in `package:platform` version 3.2 or later,
+are top-level getters, typically used like this:
```
lrhn: are instance getters of a `Platform` object, ....

pkgs/os_detect/test/fix_data_test.dart
```diff
@@ -0,0 +1,68 @@
...
+// Verifies the data-driven fixes declared in `lib/fix_data.yaml` against the
+// golden files in `test_fixes/`. See `test_fixes/README.md`.
+@TestOn('vm')
+library;
+
+import 'dart:io';
+
+import 'package:test/test.dart';
+
+/// The lowest SDK the golden files can be resolved on.
```
lrhn: SDK -> SDK version that

pkgs/os_detect/pubspec.yaml
lrhn: Should we say `^3.10.0` here.
You can't migrate to `platform` 3.2 unless you are on a 3.10 SDK anyway.
(But I guess the deprecation works, even if the migration does not.)

pkgs/os_detect/lib/fix_data.yaml
```diff
@@ -0,0 +1,153 @@
...
+  # `os_detect` in a browser should check `Platform.current.isBrowser` first.
+  - title: "Use NativePlatform.current!.operatingSystem"
+    date: 2026-09-14
+    element:
+      uris: ["package:os_detect/os_detect.dart"]
+      getter: "operatingSystem"
+    changes:
+      - kind: "replacedBy"
+        newElement:
+          uris: ["package:platform/platform.dart"]
```
lrhn: (Will this add `platform` to dependencies, or do you have to fix that afterwards? Just curious, either is fine. Probably safest to not add something to dependencies automatically, if it should only be a dev-dependency.)

pkgs/os_detect/test/fix_data_test.dart
```diff
@@ -0,0 +1,68 @@
...
+///
+/// `test_fixes/` depends on `package:platform` 3.2, which requires Dart 3.10.
+const _minimumSdk = (major: 3, minor: 10);
+
+void main() {
+  final testFixes = Directory('test_fixes');
+
+  test('fix_data.yaml matches the golden files in test_fixes/', () async {
+    final pubGet = await Process.run(
+        Platform.resolvedExecutable,
```
lrhn: Should this use Kevin's new cli-helper to find the executable?
https://pub.dev/documentation/cli_util/latest/cli_util/dartExecutable.html

(Not sure `Platform.resolvedExecutable` works if the test is AOT compiled. I don't know if that can happen, or if we should just allow it to fail then.)

pkgs/os_detect/test/fix_data_test.dart
```diff
@@ -0,0 +1,68 @@
...
+    );
+  }, skip: _skipReason(testFixes));
+}
+
+/// Why the golden comparison cannot run here, or `null` if it can.
+String? _skipReason(Directory testFixes) {
+  if (!testFixes.existsSync()) {
+    return 'Must be run from the package root; ${testFixes.path} not found.';
+  }
+  final version = Platform.version.split(' ').first.split('.');
```
lrhn: No need for the `split(' ')`. It's a SemVer version "possibly followed by whitespace and other version and build details."

That whitespace could be a non-space character, so splitting on space isn't enough.
But you also only care about the  major and minor versions, so just splitting on the first two `.`s is enough. It doesn't matter what's after the second `.`.

pkgs/os_detect/test/fix_data_test.dart
```diff
@@ -0,0 +1,68 @@
...
+  final major = int.tryParse(version.first) ?? 0;
+  final minor = version.length > 1 ? int.tryParse(version[1]) ?? 0 : 0;
```
lrhn: This does defensive parsing (may not have two elements, may not contain integer numerals).
I'd probably trust the documentation and assume it definitely starts with a SemVer version,
and allow it to crash if it doesn't.

```dart
  final version = Platform.version;
  final dot1Index = version.indexOf('.');
  final dot2Index = version.indexOf('.', dot1Index + 1);
  // Can check `dot1Index > 0 && dot2Index > dot1Index + 1` here if want to.
  final major = int.parse(version.substring(0, dot1Index);
  final minor = int.parse(version.substring(dot1Index + 1, dot2Index);
```

Or if actually  I'd make the error message reflect that it's not a valid semver version.

Consider:
```dart
  final semVerPattern(r'^(\d+)\.(\d+)\.');
  if (semVerPattern.firstMatch(Platform.version) case final match?) {
    var major = int.parse(match[1]!);
    var minor = int.parse(match[2]!);
    if (major > _minimumSdk.major ||
        major == _minimumSdk.major && minor >= _minimumSdk.minor) {
       return null;
    }
    return 'Requires ... ';
  }
  return 'Platform.version is not a valid SemVer version: ${Platform.version}';
```
 I know I don't usually recommend RegExps, but this is precisely the job they're good at: Matching a relatively simple pattern with varying length sub-patterns.

(Could validate the entire SemVer version following the second `.`, but to do that properly is a more complicated RegExp, and probably not worth it.)

pkgs/os_detect/test_fixes/os_detect.dart.expect
```diff
@@ -0,0 +1,28 @@
...
+bool get onBrowser => Platform.current.isBrowser;
+bool get onFuchsia => Platform.current.isFuchsia;
+bool get onIOS => Platform.current.isIOS;
+bool get onLinux => Platform.current.isLinux;
+bool get onMacOS => Platform.current.isMacOS;
+bool get onWindows => Platform.current.isWindows;
+
+void report() {
+  if (Platform.current.isBrowser) {
+    print(NativePlatform.current!.operatingSystemVersion);
```
lrhn: This will definitely throw in a browser, because `NativePlatform.current` is `null`.

There is no direct migration of `operatingSystemVersion` that works in a browser, because that's a native platform only property.
(That's an API improvement, you can't ask for things that don't make sense on the platform you're on.)

If you _really_ want to migrate this to work in all cases, it may have to be:
```dart
 (NativePlatform.current?.operatingSystemVersion ?? BrowserPlatform.current!.version)
```

lrhn (approved):

